PR #5, opened 2022-09-08
Initial draft of TEE and TEE AIA SBI extensions

atulkharerivos:
Here's a first pass at transcribing the existing TEE and TEE AIA interfaces into a SBI format. Note that we'll have to revisit the implementation of sbi_tee_tvm_add_measured_pages() since it's likely not compatible with RISCV32.


specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+
+=== Function: Create TVM (FID #0)
+[source, C]
+-----
+struct sbiret sbi_tee_create_tvm(unsigned long tvm_create_params,
+                                 unsigned long tvm_create_params_len);
+-----
+Creates a confidential TVM using the specified parameters. The tvm_create_parameters
+is a pointer to the `tvm_create_params` described below, and tvm_create_params_len is
+the size of the structure in bytes.
```
atishp04, line 56: Why do we need the len parameter ? tvm_create_params struct is a fixed one
.

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+Callers of this API should first invoke `sbi_tee_tsm_get_info()` to obtain information
+about the parameters that should be used to populate `tvm_create_params`.
+
+[source, C]
+----
+struct tvm_create_params {
+    /* 
+    The base physical address of the 16KiB confidential memory region
+    that should be used for the TVM’s page directory. Must be 16KiB-aligned.
+    */
```
atishp04, line 67: nit comment: The comment block should follow the standard comment styles. 

/*
 * Comment block
 * /

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+[source, C]
+-----
+struct sbiret sbi_tee_create_tvm(unsigned long tvm_create_params,
+                                 unsigned long tvm_create_params_len);
+-----
+Creates a confidential TVM using the specified parameters. The tvm_create_parameters
+is a pointer to the `tvm_create_params` described below, and tvm_create_params_len is
+the size of the structure in bytes. 
+
+Callers of this API should first invoke `sbi_tee_tsm_get_info()` to obtain information
```
atishp04, line 58: I think sbi_tee_tsm_get_info should be FID #0 so that the reader is already aware of the function. There are multiple reference to tsm_info in this function.

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+    unsigned long tvm_state_addr;
+    /*
+    The maximum number of vCPUs that will be created for this TVM. Must be
+    less than or equal to the value returned in tsm_info::tvm_max_vcpus
+    */ 
+    unsigned long tvm_num_vcpus;
+    /* The base physical address of the confidential memory region to be used
+    to hold the TVM’s vCPU state. Must be page-aligned and must be at least
+    TsmInfo::tvm_bytes_per_vcpu * tvm_num_vcpus bytes in length, rounded up
+    to the nearest multiple of 4KiB */
```
atishp04, line 83: nit: The comment style should be consistent through out.

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+----
+struct tvm_create_params {
+    /* 
+    The base physical address of the 16KiB confidential memory region
+    that should be used for the TVM’s page directory. Must be 16KiB-aligned.
+    */
+    unsigned long tvm_page_directory_addr;
+    /* 
+    The base physical address of the confidential memory region to be used
+    to hold the TVM’s global state. Must be page-aligned and the number of
```
atishp04, line 71: What is the global state of the TVM ? It is not explained anywhere.

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+    pages must be at least the value returned in tsm_info::vm_state_pages.
+    */
+    unsigned long tvm_state_addr;
+    /*
+    The maximum number of vCPUs that will be created for this TVM. Must be
+    less than or equal to the value returned in tsm_info::tvm_max_vcpus
+    */ 
+    unsigned long tvm_num_vcpus;
+    /* The base physical address of the confidential memory region to be used
+    to hold the TVM’s vCPU state. Must be page-aligned and must be at least
```
atishp04, line 81: Similar comment as above. It would be good to add some explanations about the vCPU & global state.

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+The possible error codes returned in `sbiret.error` are shown below.
+
+[#table_sbi_tee_create_tvm_errors]
+.TEE Create TVM Errors
+[cols="2,3", width=90%, align="center", options="header"]
+|===
+| Error code            | Description
+| SBI_SUCCESS           | The TVM with `tvm_guest_id` was successfully created.
+| SBI_ERR_INVALID_PARAM | The `tvm_create_params` or `tvm_create_params_len` parameters were invalid.
+|===
```
atishp04, line 100: can also return SBI_ERR_INVALID_ADDRESS if any of the above address is invalid ?

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+The possible error codes returned in `sbiret.error` are shown below.
+
+[#table_sbi_tee_tvm_add_page_table_pages_errors]
+.TEE TVM Add Page Table Pages Errors
+[cols="2,3", width=90%, align="center", options="header"]
+|===
+| Error code            | Description
+| SBI_SUCCESS           | The TVM with `tvm_guest_id` was successfully destroyed.
+| SBI_ERR_INVALID_PARAM | The `tvm_guest_id` was invalid.
+|===
```
atishp04, line 147: SBI_ERR_INVALID_ADDRESS here and all other following functions that uses an address in one of the parameter.

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+
+=== Function: TEE TVM Add Page Table Pages (FID #2)
+[source, C]
+-----
+struct sbiret sbi_tee_tvm_add_page_table_pages(unsigned long tvm_guest_id,
+                                               unsigned long base_page_address,
+                                               unsigned long num_pages);
+-----
+Adds num_pages 4KiB pages of confidential memory starting at page_addr to the
```
atishp04, line 131: Does it always have to be 4KiB ? What if the host prefers to create large page mappings in the future?

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+TVM's page-table page-pool. The tvm_guest_id is the unique identifier that was
+returned by a previous call to `sbi_tee_create_tvm()`.
```
atishp04, line 133: We can have a statement at the beginning about tvm_guest_id so that we can avoid repeating this in every function.

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+                                         unsigned long tsm_page_type,
+                                         unsigned long num_pages,
+                                         unsigned long tvm_gpa);
+-----
+Maps num_pages zero-filled pages of confidential memory starting at page_addr into the TVM's
+address space starting at `tvm_gpa`. The tvm_guest_id is the unique identifier returned by
+a previous call to `sbi_tee_create_tvm()`. The `guest_physical_address` must lie within a region
+of confidential memory created with `sbi_tee_tvm_add_confidential_memory_region()`. 
+
+Zero pages may be added at any time, and are typically demand faulted on TVM access.
```
atishp04, line 163: tvm_gpa is the pa of fault address ?

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+                                            unsigned long vcpu_register);
+-----
+Gets the value for the register identified by register in the VCPU identified
+by `tvm_vcpu_id` for the TVM specified by `tvm_guest_id`. The `tvm_guest_id is
+the unique identifier returned by a previous call to`sbi_tee_create_tvm()`.
+
+The permissible values for `tvm_vcpu_register`, and the associated semantics are listed below.
+[source, C]
+-----
+enum tvm_vcpu_register {
```
atishp04, line 604: It doesn't have to be defined twice. It would also make sense to keep sbi_tee_tvm_get_vcpu_register/sbi_tee_tvm_set_vcpu_register together.

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+[source, C]
+-----
+struct sbiret sbi_tee_tvm_vcpu_set_register(unsigned long tvm_guest_id,
+                                            unsigned long tvm_vcpu_id,
+                                            unsigned long tvm_vcpu_register,
+                                            unsigned long register_value);
+-----
+Sets the register identified by `tvm_vcpu_register` to the value specified by 
+`register_value` on the VCPU specified by `tvm_vcpu_id` on the TVM specified by 
+`tvm_guest_id`. The `tvm_guest_id is the unique identifier returned by a previous
```
abrestic-rivos, line 313: This is a bit of a mouthful. How about: "Sets the vCPU virtual register `tvm_vcpu_register` to `register_value`"? The meaning of `tvm_guest_id` / `tvm_vcpu_id` should be obvious from context, IMO.

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+[source, C]
+-----
+struct sbiret sbi_tee_tvm_add_confidential_memory_region(unsigned long tvm_guest_id,
+                                                         unsigned long tvm_gpa_addr,
+                                                         unsigned long region_len);
+-----
+Marks the range of TVM physical address space starting at `tvm_gpa_addr` as reserved
+for the mapping of confidential memory. The region is initially unpopulated, and pages
+of confidential memory may be inserted by calling `sbi_tee_tvm_add_zero_pages()` and
+`sbi_tee_tvm_add_measured_pages()`.Both `tvm_gpa_addr` and `region_len` must be 4kB-aligned.
```
abrestic-rivos, line 660: nit: Space after "."

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+.TEE TVM Add Shared Pages
+[cols="2,3", width=90%, align="center", options="header"]
+|===
+| Error code            | Description
+| SBI_SUCCESS           | The operation completed successfully
+| SBI_ERR_INVALID_PARAM | One or more parameters were invalid
+|===
+
+== TEE AIA Extension (EID #0x54414941)
+
```
abrestic-rivos, line 773: Probably should include some prose here about what this extension is for?

Also, this is still very WIP :)

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+and the `tvm_vcpu_id` must have been previously added using `sbi_tee_tvm_vcpu_create`.
+
+The `tvm_guest_id` must be in a "runnable" state. Note that the function does not
+return unless the TVM exits with a trap that cannot be handled by the TVM.
+
+[source, C]
+-----
+enum tvm_vcpu_exit_code {
+    /* The vCPU exited due to an interrupt directed at the host. */
+    host_interrupt = 0,
```
abrestic-rivos, line 215: Should enumerators be in `ALL_CAPS` to be consistent with other enums in the SBI spec?

specification/sbi_atee.adoc
atishp04: I think chronological sequence more sense in the spec as it helps coherent reading. All other SBI extensions assign FIDs in that order as well.

specification/sbi_atee.adoc
atishp04: Ahh okay. Makes sense.

specification/sbi_atee.adoc
atishp04: Yes. Do we want to restrict that option from beginning ? How about passing the page size as an argument ? 
TSM can always error if OS/VMM passes hugepages that TSM doesn't support.

specification/sbi_atee.adoc
atulkharerivos: Possibly, don't believe we return that code in our implementation, but will update.

specification/sbi_atee.adoc
atulkharerivos: I have taken a stab at it, and have added more exposition...PTAL.

specification/sbi_atee.adoc
atulkharerivos: See the exposition to see if it helps. I have removed references to the global.

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+The possible error codes returned in `sbiret.error` are shown below.
+
+[#table_sbi_tee_tvm_vcpu_create_errors]
+.TEE TVM VCPU Run Errors
+[cols="2,3", width=90%, align="center", options="header"]
+|===
+| Error code            | Description
+| SBI_SUCCESS           | The TVM exited and sbiret.value contains the tvm_exit_code
+| SBI_ERR_INVALID_PARAM | The `tvm_guest_id` or `tvm_vcpu_id` was invalid, or guest
+                          wasn't runnable.
```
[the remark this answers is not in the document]
atulkharerivos: Possibly...does it require defining a new SBI code?

specification/sbi_atee.adoc
atulkharerivos: I have removed global, and added some exposition...PTAL.

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+for actually running the payloads, like memory allocation, scheduling, interrupt handling, 
+virtual devices is left to the OS/VMM (which are outside the TCB).
+
+In this scheme, most resources like memory, (V)CPUs start off as traditional untrusted resources
+in the non-confidential world, and are transtioned to their confidential analogue via the TSM.
+Once the conversion process is complete, a confidential-resource cannot be reverted to the 
+non-confidential world unless it's is no longer assigned to a TVM.
+
+The TEE address space can be comprised of confidential and non-confidential regions. The former includes
+both measured and code pages that are part of the TVM payload, and confidential zero-pages that can be
```
abrestic-rivos, line 73: nit: Drop "and code pages" since they're measured by definition.

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+mapped in by the VMM on demand. The non-confidential regions include non-confidential memory that can
+be used for virtual device emulation and so forth.
```
abrestic-rivos, line 75: maybe: "The non-confidential regions include memory that can be shared with the VMM or to emulate MMIO"

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+TSM.
+
+. Platform TSM detection and capability enumeration
+. Conversion of non-confidential memory to confidential memory
+. Trusted VM (TVM) creation
+. Donating confidential memory to the TSM for TVM page management
+. Defining confidential memory regions
+. Defining MMIO regions
+. Defining non-confidential shared memory regions
+. Conversion of non-confidential (V)CPUs to confidential-VCPUs
```
abrestic-rivos, line 90: What is meant by "Conversion of ... (V)CPUs"?

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+case of an irrecoverable exception.
+
+Page faults and MMIO requests can be processed using `sbi_tee_tvm_add_zero_pages()`, `sbi_tee_tvm_add_shared_pages()`,
+and `sbi_tee_tvm_vcpu_set_register()`, and then calling `sbi_tee_tvm_vcpu_run()` to resume execution.
+
+=== TVM teardown
+
+The host can teardown a TVM by calling `sbi_tee_destroy_tvm()`. This automatically releases all confidential memory
+assigned to the TVM, and it can be repurposed for use with other TVMs. However, reclaiming the memory for use by
+non-confidential workloads requires and explicit call to `sbi_tee_tsm_reclaim_pages`.
```
abrestic-rivos, line 185: s/and/an/

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+-----
+Copies num_pages pages from non-confidential memory at `source_address` to confidential
+memory at `dest_addr`, then measures and maps the pages at `dest_addr` at the TVM physical
+address space at `tvm_guest_gpa. The mapping must lie within a region of confidential memory
+created with `sbi_tee_tvm_add_confidential_memory_region()`. The tsm_page_type parameter must
+be a legal value for enum type `tsm_page_type`.
+
+This call may not not be made after calling `sbi_tee_tvm_finalize()`.
+
+The possible error codes returned in `sbiret.error` are shown below.
```
abrestic-rivos, line 564: I wonder if we should have an error code for "not enough page table pages" for the various `add_*_pages` calls.

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+This requires the host to make three separate ECALLs to the TSM:
+
+. sbi_tee_tsm_convert_pages()
+. sbi_tee_tsm_initiate_fence()
+. sbi_tee_tsm_local_fence()
+
+The memory conversion process is complete when sbi_tee_tsm_local_fence() is successfully completed
+on the every VCPU on the platform.
+
+Converted memory may be assigned to TVMs, but cannot be repurposed for regular use unless it's reclaimed.
```
atishp04, line 128: Some additional clarification would be good.  
Converted memory may be assigned to other TVMs, but cannot be repurposed for non-confidential scenarios unless it's reclaimed.

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+. Tearing down TVMs
+. Repurposing confidential memory for other TVMs
+. Reclaiming confidential memory for non-confidential VMs
+
+=== Platform TSM detection and capability enumeration
+
+Platform support for the TSM can be detected by probing for the EXT_TEE extension, and then
+calling sbi_tee_tsm_get_info() to get information about the current status of the TSM. The
+TSM must be be in TSM_READY in order to process further ECALLs.
+
```
atishp04, line 109: Describe the TSM states here in a table and point to that table in tsm_get_info ?

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+on the every VCPU on the platform.
+
+Converted memory may be assigned to TVMs, but cannot be repurposed for regular use unless it's reclaimed.
+
+=== TVM creation
+
+TVMs are created using the sbi_tee_create_tvm(). This creates a TVM, and sets with state "initializing". The
+host must assign confidential memory for page tables, payload mapping, and VCPUs before it can be transitioned
+into a "runnable" state.
+
```
atishp04, line 135: Table for all states of TVM. It would be good to add a state transition diagram too. But that can be added later. Don't want to burden you with so much documentation ;)

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+. Dontations for the page-table pool
+.. The host must ensure that the TSM has sufficient confidential memory for mapping and
+   managing TVM page-tables for the code and data payloads (sbi_tee_tvm_add_page_table_pages())
+. Defining memory regions for confidential, MMIO, and shared memory regions
+.. The host must declare the TVM physical address ranges for mapping of confidential and
+   non-confidential memory. There can be multiples ranges of each type, but no two regions
+   can overlap. Both confidential and non-confidential regions can be sparsely mapped; however,
+   any sparsely mapped confidential page that's demand-page can only be a demand-zero page. 
+   The `sbi_tee_tvm_add_confidential_memory_region()`, `sbi_tee_tvm_add_emulated_mmio_region()`,
+   and `sbi_tee_tvm_add_shared_memory_region()` can be used for this purpose.
```
atishp04, line 149: I suggested earlier to combine these three into one function and pass a type parameter. Any thoughts on that ?

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+. Mapping of zero-page confidential pages to the TVM
+.. The host can map zero-pages in a TVM confidential memory range at any point of time by
+   calling `sbi_tee_tvm_add_zero_pages()`.
+. Mapping non-confidential shared pages on demand
+.. The host can share non-confidential memory with the TVM by calling `sbi_tee_tvm_add_shared_pages()`
+
+
+=== VCPU assignment
+
+The host must register platform VCPUs with the TSM before they can be used for TVM executio by calling
```
atishp04, line 162: nit: /s/executio/execution

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+Following assignment of memory and VCPU resources, the host can transition the guest into a "runnable"
+state by calling `sbi_tee_tvm_finalize()`. Note that some TEE calls are no longer permissble after this
+transition.
+
+The host can use the `sbi_tee_tvm_vcpu_set_register()` to setup TVM execution parameters like the entry
+point, and then call `sbi_tee_tvm_vcpu_run()` to begin execution. TVM execution continues until there an
+event like an interrupt, or fault that cannot be serviced by the TSM. The host can examine the return code
+from the ECALL, and use `sbi_tee_tvm_get_vcpu_register()` to determine further course of action. This may
+involve serving page-faults for demand-page regions, servicing MMIO requests, or terminating the TVM in
+case of an irrecoverable exception.
```
atishp04, line 176: We should get rid of entryPC & entryArg once we have the shared memory based approach in place to replace set/get registers.

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+
+The possible error codes returned in `sbiret.error` are shown below.
+
+[#table_sbi_tee_tsm_get_info_errors]
+.TEE TSM Get Info
+[cols="2,3", width=90%, align="center", options="header"]
+|===
+| Error code              | Description
+| SBI_SUCCESS             | The operation completed successfully.
+| SBI_ERR_INVALID_ADDRESS | An input address was invalid.
```
atishp04, line 296: tsm_info_address was invalid.

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+
+=== Function: TEE TSM Get Info (FID #0)
+[source, C]
+-----
+struct sbiret sbi_tee_tsm_get_info(unsigned long tsm_info_address,
+                                   unsigned long tsm_info_len);
+-----
+Writes up to `tsm_info_len` bytes of information the physical memory address
+specified by `tsm_info_address`. The minimum expected length for the `tsm_info_len`
+parameter is described by the `tsm_info` struct below. The information returned
```
atishp04, line 248: Bit ambiguous statement. tsm_info_len should be the size of `tsm_info` struct described below.

specification/sbi_atee.adoc
atishp04: I don't see it being resolved ?

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+| Error code            | Description
+| SBI_SUCCESS           | The operation completed successfully.
+| SBI_ERR_INVALID_PARAM | The input parameters were invalid.
+| SBI_ERR_FAILED        | The operation failed for unknown reasons.
+|===
+
+=== Function: TEE TSM Initiate Fence (FID #2)
+[source, C]
+-----
+struct sbiret sbi_tee_tsm_initiate_fence(void);
```
atishp04, line 333: Does it make more sense to call it sbi_tee_tsm_global_fence ?

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+-----
+Initiates a TLB invalidation sequence for all pages marked for conversion via
+calls to `sbi_tee_tsm_convert_pages()` between the previous `sbi_tee_tsm_initiate_fence()`
+and now. The TLB invalidation sequence is completed when `sbi_tee_tsm_local_fence()` has
```
atishp04, line 337: Not sure what do you mean by "between the previous `sbi_tee_tsm_initiate_fence()`
and now." ?

atishp04 (commented):
Overall looking good. Left mostly nit comments.
It would also be good to remove the big table with old function ID names at the end.
You can just have a table with all FIDs for both extensions.

specification/sbi_atee.adoc
atulkharerivos: Yes -- see Andrew's PR here: https://github.com/rivosinc/salus/pull/96

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
 The following operations must be implemented by the TSM ( _Formal input 
 parameters and return codes are TBD_ ).
```
rsahita, line 45: remove this TBD comment?

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
 
+== Background
+
+Virtualization platforms are typically comprised of several components including
+platform firmware, host OS, VMM, and the actual payloads that run on them (typically
+in a VM). This model is well established, but the downside is that most platform
+components are in the TCB. This aspect is ill-suited for confidential compute
+workloads (Trusted Execution Environments) that strive to minimize the TCB footprint.
```
rsahita, line 53: nit - in bracket text - "(that rely on Trusted Execution Environments)"

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+
+We seek to alleviate this concern by introducing the notion of the a trusted broker for
+TEEs known as the TEE Security Manger (TSM). The idea is that the TSM has a minimal possible
+footprint, and acts as a trusted intermediary between the the untrusted platform components,
+and the TEE workloads that run on it. This is accomplished by having a clear seperation between
+the non-confidential and confidential portions of the workloads running on the platform. The exact
+hardware and software mechanisms used to guarantee the isolation are platform dependent, but
+confidential workloads have absolute execution, data integrity, and confidentiality protections
+from amy component in the non-confidential world. The responsibility of the TSM is to enforce
```
rsahita, line 62: in the above text suggest clarify and restate as - "This is accomplished by the TCB (which includes the TSM and HW) to enforce strict confidentiality and integrity security properties for code execution and data integrity for confidential workloads. The TCB isolates a confidential workload from non-confidential workloads as well as other confidential workloads on the same platform. The exact hardware....".

also I would remove the superlatives like "absolute "

typo "amy" - clarify SW component, and some HW adversary components
from amy component in the non-confidential world

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+TEEs known as the TEE Security Manger (TSM). The idea is that the TSM has a minimal possible
+footprint, and acts as a trusted intermediary between the the untrusted platform components,
+and the TEE workloads that run on it. This is accomplished by having a clear seperation between
+the non-confidential and confidential portions of the workloads running on the platform. The exact
+hardware and software mechanisms used to guarantee the isolation are platform dependent, but
+confidential workloads have absolute execution, data integrity, and confidentiality protections
+from amy component in the non-confidential world. The responsibility of the TSM is to enforce 
+the guarantees accorded to TEE workloads. Given its minimal footprint, most of the functionality
+for actually running the payloads, like memory allocation, scheduling, interrupt handling, 
+virtual devices is left to the OS/VMM (which are outside the TCB).
```
rsahita, line 65: suggest add - "Explicitly, security invariants for confidential workloads are enforced by the TSM, as opposed to the VMM which continues to manage security for non-confidential workloads. On the other hand, resource management functions for all workloads (confidential and non-confidential) are managed by the untrusted OS/VMM.".

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+
+In this scheme, most resources like memory, (V)CPUs start off as traditional untrusted resources
+in the non-confidential world, and are transtioned to their confidential analogue via the TSM.
+Once the conversion process is complete, a confidential-resource cannot be reverted to the 
+non-confidential world unless it's is no longer assigned to a TVM.
```
rsahita, line 70: add a sentence "...via the TSM. Resources once transitioned to a confidential state, may be assigned to the TVM via the TSM."

and remove double negative to restate - "Once the conversion process is complete, a confidential resource may be reclaimed to the non-confidential world only when it is not assigned to a TVM. The assignment tracking is performed by the TSM."

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+
+The TEE address space can be comprised of confidential and non-confidential regions. The former includes
+both measured that are part of the TVM payload, and confidential zero-pages that can be mapped in by the
```
rsahita, line 73: restate - "both measured pages (that are part of the initial TVM payload), and confidential (zero) pages that can be mapped in by the VMM when accessed by the TVM on demand."

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+. Platform TSM detection and capability enumeration
+. Conversion of non-confidential memory to confidential memory
+. Trusted VM (TVM) creation
+. Donating confidential memory to the TSM for TVM page management
+. Defining confidential memory regions
+. Defining MMIO regions
+. Defining non-confidential shared memory regions
+. Adding VCPUs to TVMs
+. Mapping code and data payloads for the TVM to confidential-memory regions
+. Mapping demand-zero confidential memory regions
```
rsahita, line 92: should this one be here - it is listed again below "Mapping of confidential zero pages on demand"

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+. Adding confidential VCPUs to the TVM
+. Finalizing TVM creation
+. Scheduling TVM execution
+. Management of secure interrupts
+. Handling and servicing TVM faults and exits
+. Mapping of confidential zero pages on demand
+. Mapping non-confidential shared pages on demand
+. Tearing down TVMs
+. Repurposing confidential memory for other TVMs
```
rsahita, line 101: Repurposing -> Re-assignment of

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+.TSM States
+[%header,%autowidth]
+|===
+| TSM State          | Meaning
+
+| TSM_NOT_LOADED     | TSM has not been loaded on this platform.
+| TSM_LOADED         | TSM has been loaded, but has not yet been initialized.
+| TSM_READY          | TSM has been loaded & initialized, and is ready to accept ECALLs.
+|===
+
```
rsahita, line 122: should we put in a TBD comment that states that runtime updates to the TSM will be covered in the future.

rsahita: and that may add a few new states of this table.

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+
+=== Conversion of non-confidential memory to confidential memory
+
+Platform memory is non-confidential by default, and must be converted to confidential memory
+before use with TVMs. The conversion process is initiated by designating the host physical
+pages that to be converted, and then issuing fence operations to ensure that all outstanding
+TLB entries to the non-confidential memory are flushed across all VCPUs on the platform. This
```
rsahita, line 129: across all CPUs/harts on the platform

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+regions on the platform.
+
+This requires the host to make three separate ECALLs to the TSM:
+
+. sbi_tee_tsm_convert_pages()
+. sbi_tee_tsm_global_fence()
+. sbi_tee_tsm_local_fence()
+
+The memory conversion process is complete when sbi_tee_tsm_local_fence() is successfully completed
+on the every VCPU on the platform.
```
rsahita, line 140: on every CPU on the platform.

specification/sbi_atee.adoc
rsahita: +1 (also should the TVM states be enumerated as all caps like the TSM states - in this case, TVM_INITIALIZING, TVM_RUNNABLE?

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+.. The host must ensure that the TSM has sufficient confidential memory for mapping and
+   managing TVM page-tables for the code and data payloads (sbi_tee_tvm_add_page_table_pages())
+. Defining memory regions for confidential, MMIO, and shared memory regions
+.. The host must declare the TVM physical address ranges for mapping of confidential and
+   non-confidential memory. There can be multiples ranges of each type, but no two regions
+   can overlap. Both confidential and non-confidential regions can be sparsely mapped; however,
+   any sparsely mapped confidential page that's demand-page can only be a demand-zero page. 
+   The `sbi_tee_tvm_memory_region()` can be used for this purpose.
+. Mapping of code and data payload to the TVM
+.. The host must copy the TVM code and data payload from non-confidential memory to confidential
```
rsahita, line 165: The host cannot copy into confidential memory (I know you meant the TSM does that when sbi_tee_tvm_add... is invoked) - should clarify

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+   memory by calling `sbi_tee_tvm_add_measured_pages()`
+. Mapping of zero-page confidential pages to the TVM
+.. The host can map zero-pages in a TVM confidential memory range at any point of time by
+   calling `sbi_tee_tvm_add_zero_pages()`.
```
rsahita, line 169: should add - the confidential memory page should not already be mapped, i.e. demand zero confidential pages may only be mapped for previously un-mapped pages.

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+. Mapping non-confidential shared pages on demand
+.. The host can share non-confidential memory with the TVM by calling `sbi_tee_tvm_add_shared_pages()`
+
+
+=== VCPU assignment
+
+The host must register platform VCPUs with the TSM before they can be used for TVM execution by calling
```
rsahita, line 176: I don't understand the "platform VCPUs" term - these are VCPUs specific to the TVM they are assigned to right?

specification/sbi_atee.adoc
rsahita: There was also some discussion to make finalize an implicit op as part of the first "vcpu run"

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+     * Value used to complete an emulated MMIO load by a TVM CPU. Read-write, and only
+     * accessible after the TVM has started.
+     */
+    MMIO_LOAD_VALUE = 4,
+    /*
+     * Value stored by a TVM CPU to an emulated MMIO region. Read-only, and only accessible
+     * after the TVM has started.
+     */
+    MMIO_STORE_VALUE = 5,
+};
```
rsahita, line 251: so this structure is going to change right?

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+-----
+
+=== Function: TEE TSM Get Info (FID #0)
+[source, C]
+-----
+struct sbiret sbi_tee_tsm_get_info(unsigned long tsm_info_address,
+                                   unsigned long tsm_info_len);
+-----
+Writes up to `tsm_info_len` bytes of information the physical memory address
```
rsahita, line 260: missing "at" - at the physical memory address ...

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+                                               unsigned long num_pages);
+-----
+Adds num_pages 4KiB pages of confidential memory starting at page_addr to the
+TVM's page-table page-pool. The tvm_guest_id is the unique identifier that was
+returned by a previous call to `sbi_tee_create_tvm()`. 
+
+Callers can invoke the API multiple times to add pages incrementally, but not after
+`sbi_tee_tvm_finalize()` is called.
+
+The possible error codes returned in `sbiret.error` are shown below.
```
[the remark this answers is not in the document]
rsahita: wouldnt the host know already based on its pfn db? so currently do we convert before every assignment? (which would be expensive)

specification/sbi_atee.adoc
rsahita: should we report the error case where some memory was already converted?

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+about the parameters that should be used to populate `tvm_create_params`.
+
+[source, C]
+----
+struct tvm_create_params {
+    /* 
+     * The base physical address of the 16KiB confidential memory region
+     * that should be used for the TVM’s page directory. Must be 16KiB-aligned.
+     */
+    unsigned long tvm_page_directory_addr;
```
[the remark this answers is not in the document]
abrestic-rivos: This is unavoidable, unfortunately. The H-extensions mandate a 16kB-aligned & contiguous root page table for 2nd-stage translation; see 8.5.1 in the priv spec.

specification/sbi_atee.adoc
atulkharerivos: Sure -- should this be just an "invalid parameter" or do we want to define a more specific error?

specification/sbi_atee.adoc
atulkharerivos: Should we also add an API to enumerate the TVM state?

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,936 @@ running a VMM).
...
+
+The possible error codes returned in `sbiret.error` are shown below.
+
+[#table_sbi_tee_tsm_global_fence_errors]
+.TEE TSM Initiate Fence
+[cols="2,3", width=90%, align="center", options="header"]
+|===
+| Error code            | Description
+| SBI_SUCCESS           | The operation completed successfully.
+| SBI_ERR_INVALID_PARAM | A fence operation is already in progress.
```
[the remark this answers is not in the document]
atulkharerivos: Good point -- does SBI_ERR_ALREADY_STARTED fit the bill?

specification/sbi_atee.adoc
atulkharerivos: Right...will wait for PR to be finalized.

specification/sbi_atee.adoc
rsahita: probably a NTH

atishp04 (approved):

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1238 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+                                                 unsigned long region_len);
+-----
+Marks the range of TVM physical address space starting at `tvm_gpa_addr` as reserved
+for the mapping of memory. The type of memory is specified by `tvm_memory_region_type`
+and the length is specified by by `region_len`. `tvm_memory_region_type` must be a
+legal value for the `tvm_memory_region_type` enum type described below.
+
+Both `tvm_gpa_addr` and `region_len` must be 4kB-aligned, and the region must not
+overlap with a previously defined region. This call must not be made after calling
+`sbi_tee_host_finalize_tvm()`.
```
rsahita, line 710: do we need this - once we add support for dynamic memory conversion?

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1238 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+                                              unsigned long num_pages,
+                                              unsigned long tvm_base_page_address);
+-----
+Maps num_pages zero-filled pages of confidential memory starting at `base_page_address`
+into the TVM's physical address space starting at `tvm_base_page_address`. 
+The `tvm_base_page_address` must lie within a region of confidential memory created with
+`sbi_tee_tvm_add_confidential_memory_region()`. The `tsm_page_type` parameter must be a
+legal value for the `tsm_page_type` enum.
+
+Zero pages may be added at any time, and are typically demand faulted on TVM access.
```
rsahita, line 826: but only for not-present mappings for the TVM gpa specified.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1238 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+    /* The TVM encountered a user-mode interrupt (informational only; non-resumable). */
+    USER_EXTERNAL = 8,
+    /* The TVM encountered a supervisor-mode interrupt (informational only; non-resumable). */
+    SUPERVISOR_EXTERNAL = 9,
+    /* The TVM encountered a host-mode interrupt (informational only; non-resumable). */
+    VIRTUAL_SUPERVISOR_EXTERNAL = 10,
+    /* The TVM encountered a machine-mode interrupt (informational only; non-resumable). */
+    MACHINE_EXTERNAL = 11,
+    /* The TVM encountered a supervisor-mode interrupt (informational only; non-resumable). */
+    SUPERVISOR_GUEST_EXTERNAl = 12,
```
rsahita, line 1027: general Q: whats the difference between supervisor guest external interrupt and VS_External interrupt?

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1238 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+    /* The TVM made an ECALL not directed at the host (informational only; non-resumable). */
+    MACHINE_ENVCALL = 11,
+    /* The TVM encountered an instruction page fault (informational only; non-resumable). */
+    INSTRUCTION_PAGE_FAULT = 12,
+    /* The TVM encountered a unservicable load page fault (informational only; non-resumable). */
+    LOAD_PAGE_FAULT = 13,
+    /* The TVM encountered a unservicable store page fault (informational only; non-resumable). */
+    STORE_PAGE_FAULT = 15,
+    /* The TVM made an ECALL not directed at the host (informational only; non-resumable). */
+    GUEST_INSTRUCTION_PAGE_FAULT = 20,
```
rsahita, line 1069: comment is wrong

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1238 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+     * can then be used to determine the type of memory region, and making the appropriate
+     * call (example: sbi_tee_host_add_tvm_zero_pages() to add a demand-zero confidential
+     * page if applicable), and then calling sbi_tee_host_run_tvm_vcpu to resume execution at
+     * the following instruction. 
+    */
+    GUEST_LOAD_PAGE_FAULT = 21,
+    /* 
+     * The TVM executed an instruction that caused an exit. The host should decode the
+     * instruction by examining `stval` from `tvm_vcpu_supervisor_csrs`, and determine
+     * the further course of action, and calling then calling calling sbi_tee_host_run_tvm_vcpu
```
rsahita, line 1084: typo

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1238 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+                                                           unsigned long tvm_vcpu_id,
+                                                           unsigned long tvm_vcpu_imsic_gpa);
+-------
+
+Sets the guest physical address of the specified VCPU’s virtualized IMSIC to `tvm_vcpu_imsic_gpa`.
+The `tvm_vcpu_imsic_gpa` must be valid for the AIA configuration that was set by 
+`sbi_tee_interrupt_init_tvm_aia()`. No two VCPUs may share the same `tvm_vcpu_imsic_gpa`.
+
+This can be called only after `sbi_tee_interrupt_init_tvm_aia()` and before `sbi_tee_host_finalize_tvm()`.
+All VCPUs in an AIA-enabled TVM must their IMSIC configuration set prior to calling
```
rsahita, line 1182: missing word = All VCPUs in an AIA-enabled TVM must *have* their IMSIC configuration set prior...

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1238 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+                            the TVM wasn't in the "initializing" state.
+| SBI_ERR_FAILED          | The operation failed for unknown reasons.
+|===
+
+=== Function: TEE Interrupt Convert AIA IMSIC (FID #2)
+[source, C]
+-------
+struct sbiret sbi_tee_interrupt_convert_aia_imsic(unsigned long imsic_page_addr);
+-------
+
```
rsahita, line 1204: presume this is for the MRIF - why do we treat it as special?

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1238 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+usecase for this extension is to relay information to the host.
+
+=== Function: TEE Guest Add Memory Region (FID #0)
+[source, C]
+-------
+struct sbiret sbi_tee_guest_add_memory_region(unsigned long tvm_memory_region_type,
+                                              unsigned long tvm_gpa_addr,
+                                              unsigned long region_len);
+-------
+Marks the range of TVM physical address space starting at `tvm_gpa_addr` as reserved for the mapping of
```
rsahita, line 1254: update per dynamic memory conversion

specification/sbi_atee.adoc
```diff
@@ -227,14 +1486,6 @@ page is not-present in the second stage page table and not tracked as a TEE
...
 page.                            
 
-| teecall_tg_page_accept       | Intrinsic used by the TVM to accept pages 
-as part of initial (pre-init) load or post-init add/conversion of memory 
-from non-confidential to confidential.The page mapping stays in a pending 
-state (where TLB entries cannot be created) until this operation succeeds. 
-The TSM enforces the page is zero-cleared during the page_post_add 
-performed by the VMM when the page is (re)added back as a confidential TVM 
-page.  
-
```
rsahita, line 237: we should add this back as part of the dynamic shared memory conversion support.

specification/sbi_atee.adoc
atulkharerivos: Looks like I somehow missed the conversation around dynamic memory (there probably wasn't one, except for Andrew's note about the document in the status last week). Let me take a look.

specification/sbi_atee.adoc
atulkharerivos: Sorry, not familiar with the NTH TLA :).

specification/sbi_atee.adoc
atulkharerivos: Seems like the document calls for the API to be retained? cc: @abrestic-rivos 

```
_TEE-H changes:
teeh_add_mem_region(): Same as existing behavior except that the guest may later convert parts of the region to shared, thus the host should call this for the full GPA range of the host that may be populated. The ranges added here should be in the “memory” node of the TVM’s device tree._
```

specification/sbi_atee.adoc
atulkharerivos: Good question -- this is from the Salus code -- @abrestic-rivos might know for sure.

specification/sbi_atee.adoc
atulkharerivos: @abrestic-rivos What's the difference between a GUEST_INSTRUCTION_PAGE_FAULT and INSTRUCTION_PAGE_FAULT?

specification/sbi_atee.adoc
atulkharerivos: Can you elaborate on the question? We have to convert it to confidential memory for use by the TVM, right? cc: @abrestic-rivos

specification/sbi_atee.adoc
atulkharerivos: Changed...PTAL...we need to clarify some of the interactions with sbi_tee_guest_add_memory_region() and also MMIO...please see the discussion in the document.

specification/sbi_atee.adoc
atulkharerivos: What was the API (sorry, can't see the deleted change for some reason)?

specification/sbi_atee.adoc
rsahita: nice to have

specification/sbi_atee.adoc
atishp04: Not sure if this was answered as there is a github issue for TEE specific error code. A more specific error would be helpful here as well.

specification/sbi_atee.adoc
atishp04: @rsahita Do you mean TSM should automatically add the entire memory nodes specified in the guest DT ?

atishp04: @atulkharerivos : It probably doesn't matter much as this API may change. But the current function signature doesn't match the implementation. 

Implementation:
TvmAddMemoryRegion {
        /// a0 = guest id
        guest_id: u64,
        /// a1 = type of memory region
        region_type: TeeMemoryRegion,
        /// a2 = start of the region
        guest_addr: u64,
        /// a3 = length of the region
        len: u64,
    },

specification/sbi_atee.adoc
abrestic-rivos: I think this can be removed now?

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1358 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+-----------------
+struct tvm_vcpu_register_set_location {
+    /*
+     * A value of enum type `vcpu_register_set_id`.
+     */
+    unsigned short id;
+    /*
+     * The offset of the register set from the start of the VCPU's shared-memory state area.
+     */
+    unsigned short offset;
```
abrestic-rivos, line 402: I would use the explicitly sized type (uint16_t) here.

specification/sbi_atee.adoc
abrestic-rivos: Not resovled?

specification/sbi_atee.adoc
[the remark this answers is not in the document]
abrestic-rivos: It's easy for the host to know how many PT pages have been donated in total (and it can be done in batches, not before every assignment), but it's a little more complicated for it to figure out how many of the PT pages are in use since it would require the host to track when an intermediate table has been filled for the first time. The return value here would be a shortcut for the latter and would give an indication to the host that it needs to refill the pool. (FWIW, this is roughly the approach pKVM has taken.)

specification/sbi_atee.adoc
abrestic-rivos: This will eventually become:

```
sbi_teeh_add_memory_region(tvm_id, tvm_addr, len)
```

and will only add confidential memory regions. Shared/MMIO are initiated by the guest.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1358 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+| SBI_ERR_FAILED                | The operation failed for unknown reasons.
+|===
+
+=== Function: TEE Host Create TVM VCPU (FID #15)
+[source, C]
+-----
+struct sbiret sbi_tee_host_create_tvm_vcpu(unsigned long tvm_guest_id,
+                                           unsigned long tvm_vcpu_id,
+                                           unsigned long tvm_vcpu_shared_page_addr);
+-----
```
abrestic-rivos, line 936: Update this after the changes made in https://github.com/rivosinc/salus/pull/132 ?

specification/sbi_atee.adoc
abrestic-rivos: Virtual supervisor external interrupt: an external interrupt directed at VS level.
Supervisor guest external interrupt: an HS-level interrupt indicating that are pending VS external interrupts.

The second is necessary because only one guest interrupt file may be active at a time, and the host still needs to know if an interrupt is received for one of the inactive interrupt files. See the privileged and AIA specs for more details.

BTW -- these exception/interrupt causes are simply the ones defined in the privileged spec. Not sure it's worth repeating them here. We should just refer to the privileged spec instead.

specification/sbi_atee.adoc
abrestic-rivos: Guest instruction page fault: 2nd-stage page fault on instruction fetch
Instruction page fault: 1st-stage page fault on instruction fetch

See comment above re: removing these and pointing at the privileged spec instead.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1358 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+     *
+     * |xxxxxx|Group Index|xxxxxxxxxxx|Hart Index|Guest Index| 0 |
+     *
+     * The base address is the address of the IMSIC with group ID, hart ID, and guest ID of 0.
+     */
+    unsigned long imsic_base_addr;
+    /* The number of group index bits in an IMSIC address. */
+    unsigned int group_index_bits;
+    /* The location of the group index in an IMSIC address. Must be >= 24. */
+    unsigned int group_index_shift;
```
abrestic-rivos, line 1142: General comment (and maybe @atishp04 has an opinion here too): I feel like we should try to use explicitly-typed sizes in in-memory structures that cross the ABI boundary, with a possible exception of using `unsigned long` for pointer types (though 64-bit would still be fine for RV32 since physical addresses are 34-bits there).

specification/sbi_atee.adoc
abrestic-rivos: This is not an MRIF -- this is an actual guest interrupt file page of the IMSIC and the actual HW mechanism of converting the IMSIC page from non-conf to conf is implementation-defined, hence the separate API.

specification/sbi_atee.adoc
abrestic-rivos: (this is tg_accept, FYI)

We should continue the discussion of shared conversion on the doc; I'm not sure we need to introduce guest memory acceptance just for the purposes of memory un-sharing.

specification/sbi_atee.adoc
atulkharerivos: Changed to uint32_t to match the previous uint16_t from above, but will defer to @atishp04 if unsigned int / unsigned short is preferred.

specification/sbi_atee.adoc
dgreid: We should finish the discussion about sharing API with nested VMs before comitting to _only_ confidential memory support here. Could handle that with a separate API for adding shared memory, but that feels like a step backwards.

abrestic-rivos: Is there a good reason to even bother with memory regions for non-TEE nested virtualization though? The guest hypervisor should be allowed to insert pages or handle page faults in whatever way it wants for non-TVMs; requiring it to declare valid GPA ranges up front imposes a TEE-specific requirement that wouldn't otherwise be present in the non-nested case.

dgreid: Good point. As long as we have a path that allows mapping outside of a pre-registered region that should work fine.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1299 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+|===
+
+=== Function: TEE Host Add TVM Page Table Pages (FID #9)
+[source, C]
+-----
+struct sbiret sbi_tee_host_add_tvm_page_table_pages(unsigned long tvm_guest_id,
+                                                    unsigned long base_page_address,
+                                                    unsigned long tsm_page_type,
+                                                    unsigned long num_pages);
+-----
```
atishp04, line 719: This is incorrect. 
AddPageTablePages {
                guest_id,
                page_addr,
                num_pages,
            } => self
                .guest_add_page_table_pages(guest_id, page_addr, num_pages)
                .into(),

atulkharerivos: Actually, the tsm_page_type parameter was added per your suggestion :). We do need to reconcile it the Salus implementation...I will submit a PR if the tsm_page_type is still desirable.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1299 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+};
+
+struct tsm_info {
+    /* 
+     * The current state of the TSM. If the state is not TSM_READY, the remaining
+     * fields are invalid and will be initialized to 0.
+     */
+    enum tsm_state tsm_state;
+    /* Version number of the running TSM. */
+    unsigned long tsm_version;
```
atishp04, line 404: In Salus, it is defined as u32.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1299 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+struct sbiret sbi_tee_host_add_tvm_page_table_pages(unsigned long tvm_guest_id,
+                                                    unsigned long base_page_address,
+                                                    unsigned long tsm_page_type,
+                                                    unsigned long num_pages);
+-----
+Adds num_pages confidential memory of type `tsm_page_type` starting at `base_page_address` to the
+TVM's page-table page-pool. `tsm_page_type` must a legal value for the `tsm_page_type` enum. Note
+that the TSM may not support all page types.
+
+This call must not be made after calling `sbi_tee_host_finalize_tvm()`.
```
atishp04, line 724: This can be called after sbi_tee_host_finalize_tvm. Ideally, we want to add most of the page table pages on fault itself.

specification/sbi_atee.adoc
atishp04: yeah. That is better. I already tripped once on this ;)

specification/sbi_atee.adoc
atishp04: As discussed, explicit size for tsm_state as well.

specification/sbi_atee.adoc
```diff
@@ -13,6 +43,858 @@ running a VMM).
...
+[source, C]
+-----
+struct sbiret sbi_tee_tvm_vcpu_run(unsigned long tvm_guest_id,
+                                   unsigned long tvm_vcpu_id);
+-----
+Runs the VCPU specified by `tvm_vcpu_id` in the TVM specified by `tvm_guest_id`.
+The `tvm_guest_id` is the unique identifier returned by a previous call to`sbi_tee_create_tvm()`,
+and the `tvm_vcpu_id` must have been previously added using `sbi_tee_tvm_vcpu_create`.
+
+The `tvm_guest_id` must be in a "runnable" state. Note that the function does not
```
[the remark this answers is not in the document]
atulkharerivos: @atishp04 Do we need anything beyond the following?

enum tvm_state {
    /* The TVM has been created, but isn't yet ready to run */
    TVM_INITIALIZING = 0,
    /* The TVM is in a runnable state */
    TVM_RUNNABLE = 1,
};

specification/sbi_atee.adoc
atulkharerivos: @abrestic-rivos I had added the following blurb...do we need more?
```
The TEE Interrupt extension supplements the TEE Host extension with hardware-assisted interrupt
virtualization using the RISC-V Advanced Interrupt Architecture (AIA) on platforms which
support it.
```

specification/sbi_atee.adoc
```diff
@@ -13,6 +44,953 @@ running a VMM).
...
+
+The possible error codes returned in `sbiret.error` are shown below.
+
+[#table_sbi_tee_tvm_add_measured_pages_errors]
+.TEE TVM Add Measured Pages
+[cols="2,3", width=90%, align="center", options="header"]
+|===
+| Error code              | Description
+| SBI_SUCCESS             | The operation completed successfully.
+| SBI_ERR_INVALID_ADDRESS | An input address was invalid.
```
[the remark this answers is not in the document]
atulkharerivos: Changed

specification/sbi_atee.adoc
atulkharerivos: @rsahita Believe this is resolved...can you PTAL?

rsahita (changes requested):
thanks Atul - these flows are very good to understand the ABI interactions - some comments:

On the tsm_detection_and_tvm_creation flow: 

*as part of the add vcpu loop - TSM should report the number of vcpus that have been added- if the relying party does not see the expected number it can disapprove attestation.
*should also show the other vcpus being entered - since the guest OS will typically maintain its own checks to register all TVM (v)cpus

On the tvm_runtime_execution flow: 

* on the fault/ecall exit - should note that TSM saves (and scrubs) TVM GPR and extended register state (V) before notifying host about TVM exit - similarly loads vhart saved state to resume
* for demand fault in of confidential memory - should show that the TVM has to accept the GPA before it can use the mapping. 
* we should show a flow for secure interrupt handling - as well as debug and perfmon opt-in also in this runtime svg

On the tvm_destruction_and_memory_reclamation flow:

* should make a top level note that the host may stop a TVM unilaterally (it does not have to wait for a TVM to exit from TVM-execution loop) - it can achieve that via injection of an interrupt which causes a trap into the TSM giving control to the VMM. (so should show an IPI sequence)
* post the destroy TVM call - to reclaim confidential memory, should show the same sequence as before to synchronize TLBs to prevent stale mappings being held in a hart for a TVM (and to flush any tracking information that may being held in caches).

rsahita (approved):

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1296 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+
+[#table_sbi_tee_host_add_tvm_page_table_pages_errors]
+.TEE Host Add TVM Page Table Pages
+[cols="2,3", width=90%, align="center", options="header"]
+|===
+| Error code              | Description
+| SBI_SUCCESS             | The operation completed successfully.
+| SBI_ERR_INVALID_ADDRESS | `base_page_address` was invalid.
+| SBI_ERR_INVALID_PARAM   | `tvm_guest_id` or `num_pages` were invalid,
+                             or `tsm_page_type` is invalid.
```
rsahita, line 732: | SBI_ERR_OUT_OF_PTPAGES | The operation could not complete due to insufficient page table pages.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1296 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+in the non-confidential world, and are transtioned to their confidential analogue via the TSM.
+Once the conversion process is complete, it can be assigned to a TVM via the TSM. A converted confidential-resource
+can be freely assigned to another TVM when it's no longer in use. However, an unused confidential-resource must be
+explicitly reclaimed for use in the non-confidential world (this is tracked and enforced by the TSM).
+
+The TEE address space can be comprised of confidential and non-confidential regions. The former includes
+both measured pages (that are part of the initial TVM payload), and confidential zero-pages that can be
+mapped-in on demand by the VMM following runtime access by the TVM. The non-confidential TVM-defined regions
+include those for shared-pages and MMIO.
+
```
rsahita, line 71: we should note here MMIO is non-confidential with no direct IO. once direct IO is supported - MMIO page mappings will also be allowed to be in the confidential memory region (simply since the mappings are in confidential memory).

specification/sbi_atee.adoc
rsahita: @abrestic-rivos was this updated here? do we need to revert the deleted lines above?

rsahita (approved):

specification/sbi_atee.adoc
rsahita: resolved - we used the guest-side blocking TLB flush approach instead of accept.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1296 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+| SBI_ERR_INVALID_ADDRESS | `base_page_address` was invalid.
+| SBI_ERR_INVALID_PARAM   | `num_pages` was invalid.
+| SBI_ERR_FAILED          | The operation failed for unknown reasons.
+|===
+
+=== Function: TEE Host Initiate Global Fence (FID #3)
+[source, C]
+-----
+struct sbiret sbi_tee_host_global_fence(void);
+-----
```
abrestic-rivos, line 505: nit: I would drop the "global" here. It's confusing in the presence of local_fence below.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1296 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+|===
+
+=== Function: TEE Host Add TVM Memory Region (FID #8)
+[source, C]
+-----
+struct sbiret sbi_tee_host_add_tvm_memory_region(unsigned long tvm_guest_id,
+                                                 unsigned long tvm_memory_region_type,
+                                                 unsigned long tvm_gpa_addr,
+                                                 unsigned long region_len);
+-----
```
abrestic-rivos, line 663: `tvm_add_memory_region` has changed to only allow confidential memory regions, so the `tvm_memory_region_type` has been dropped. The text below needs to be updated as well.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1296 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+                                              unsigned long num_pages,
+                                              unsigned long tvm_base_page_address);
+-----
+Maps num_pages zero-filled pages of confidential memory starting at `base_page_address`
+into the TVM's physical address space starting at `tvm_base_page_address`. 
+The `tvm_base_page_address` must lie within a region of confidential memory created with
+`sbi_tee_host_add_tvm_memory_region()`. The `tsm_page_type` parameter must be a
+legal value for the `tsm_page_type` enum.
+
+Zero pages for non-present TVM-specified GPA ranges may be added at any time, and are typically demand faulted on TVM access.
```
abrestic-rivos, line 786: Zero pages can only be added post-finalization.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1296 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+course of action (if sbiret.value is 0).
+
+The TSM sets the most significant bit in `scause` to indicate that that the exit was caused
+by an interrupt, and if this bit is clear, the implication is that the the exit was caused
+by an exception. The remaining bits specific information about the interrupt or exception,
+and the specific reason can be determined using the enumeration detailed below.
+
+[source, C]
+-------
+enum tvm_interrupt_exit {
```
abrestic-rivos, line 964: Can we just point to the priv spec instead of defining an enum that matches exactly the constants defined in the priv spec?

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1296 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+     * can then be used to determine the type of memory region, and making the appropriate
+     * call (example: sbi_tee_host_add_tvm_zero_pages() to add a demand-zero confidential
+     * page if applicable), and then calling sbi_tee_host_run_tvm_vcpu to resume execution at
+     * the following instruction.
+     */
+    GUEST_STORE_PAGE_FAULT = 23,
+};
+-------
+
+== TEE Interrupt Extension (EID #0x54414949)
```
abrestic-rivos, line 1040: Needs to be updated for interrupt injection.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1296 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+The possible error codes returned in `sbiret.error` are shown below.
+
+[#table_sbi_tee_guest_local_fence_errors]
+.TEE Guest Local Fence
+[cols="2,3", width=90%, align="center", options="header"]
+|===
+| Error code            | Description
+| SBI_SUCCESS           | The operation completed successfully.
+| SBI_ERR_FAILED        | The operation failed for unknown reasons.
+|===
```
abrestic-rivos, line 1280: Remove these.

specification/sbi_atee.adoc
abrestic-rivos: `add_memory_region` has been replaced by `add_mmio_region`. Sharing/unsharing is handled with separate APIs.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1296 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+[source, C]
+-------
+struct sbiret sbi_tee_guest_share_memory_region(unsigned long tvm_gpa_addr,
+                                                unsigned long region_len);
+-------
+Marks the range of TVM physical address space starting at `tvm_gpa_addr` as reserved for the mapping of non-confidential memory. The entire range must reside in a region of
+type `CONFIDENTIAL_REGION`, and all pages must have been previously unmapped and flushed (using `sbi_tee_guest_flush_pages()`, followed by a fence).
+
+Both `tvm_gpa_addr` and `region_len` must be 4kB-aligned. This call will result in an exit to the
+host on success. Following successful completion, the TVM can access all any page in the range.
```
abrestic-rivos, line 1292: This needs to be re-written to reflect what was actually implemented. I'd suggest something like this:

"""
Converts the requested range of TVM guest physical address space from confidential to shared. The requested range must lie within an existing region of confidential address space, and may or may not be populated.

If the region of address space is populated, the TSM invalidates the pages and marks the region as pending conversion to shared. The host must complete a TVM TLB invalidation sequence, initiated by `tee_host_tvm_initiate_fence()`, in order to complete the conversion. The calling TVM vCPU is considered blocked until the conversion is completed; attempts to run it with `tee_host_tvm_run()` will fail with <some error code>. Any guest page faults taken by other TVM vCPUs in this region prior to completion of the conversion are considered fatal. The host may not insert any pages in the region prior to completion of the conversion. Upon completion, the host may reclaim the confidential pages that were previously mapped in the region using `tee_host_tsm_reclaim_pages()` and may insert shared pages into the region using `tee_host_tvm_add_shared_pages()`.

If the range of address space is completely unpopulated, the region is immediately converted to shared and the host may insert shared pages.
"""

cc: @rsahita

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1296 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+-------
+struct sbiret sbi_tee_guest_unshare_memory_region(unsigned long tvm_gpa_addr,
+                                                  unsigned long region_len);
+-------
+Marks the range of TVM physical address space starting at `tvm_gpa_addr` as reserved for the mapping of confidential memory (the contents will be guaranteed to be zeroes on the first
+access). The entire range must reside in a region of type `SHARED_MEMORY_REGION`, and all
+pages must have been previously unmapped and flushed (using `sbi_tee_guest_flush_pages()`, followed by a fence).
+
+Both `tvm_gpa_addr` and `region_len` must be 4kB-aligned. This call will result in an exit to the
+host on success. Following successful completion, the TVM can access any page in the range.
```
abrestic-rivos, line 1318: Same as above, but swap shared/confidential.

specification/sbi_atee.adoc
```diff
@@ -10,8 +39,1296 @@ shown in the figure above, the TEE Runtime (operating in VS-mode) supports
...
+| Error code              | Description
+| SBI_SUCCESS             | The operation completed successfully.
+| SBI_ERR_INVALID_ADDRESS | `imsic_page_addr` was invalid.
+| SBI_ERR_INVALID_PARAM   | The memory is still assigned to a TVM.
+| SBI_ERR_FAILED          | The operation failed for unknown reasons.
+|===
+
+== TEE Guest Extension (EID 0x54454547)
+The TEE Guest extension supplements the TEE Host extension, and TVMs to communicate with TSM. A typical
+usecase for this extension is to relay information to the host.
```
abrestic-rivos, line 1183: Probably need a blurb here about how TEE-Guest calls cause an exit to the host with scause set to ECALL, a6 set to FID, a0-a5 set to ECALL args, etc.

specification/sbi_atee.adoc
rsahita: this LGTM - I can merge this change into the writeup.

specification/sbi_atee.adoc
rsahita: ditto

rsahita (approved):

PR #7, opened 2022-12-01
Add gem mathematical.so build dependencies in 22.04

jjscheel:
Signed-off-by: Jeff Scheel <jeff@riscv.org>


jjscheel:
@rsahita, this PR will fix your build issues.  That's all it does.

rsahita (approved):

PR #10, opened 2023-02-21
Minor formatting changes and csr name fixes

rajnesh-kanwal:


atishp04 (approved):

atishp04 (approved):

specification/refarch.adoc
```diff
@@ -127,7 +127,7 @@ provided as evidence of a specific TSM being loaded on a specific platform.
 During initialization, the TSM-driver will initialize a TSM-data region 
 within the TSM-memory region. The TSM-data region may hold per-hart TSM 
 state, memory assignment tracking structures and additional global data for 
-TSM management. The TSM-data region is TEE-capable memory that is apriori 
+TSM management. The TSM-data region is TEE-capable memory that is a prior
```
rsahita, line 130: the original text was ok

specification/threatmodel.adoc
```diff
@@ -57,7 +57,7 @@ in-scope adversaries
 T6: Incorrect execution of workload via runtime modification of CPU 
 registers, CSRs, mode switches via in-scope adversaries
 
-T7: Invalid code execution or data injection/replacement via second stagel 
+T7: Invalid code execution or data injection/replacement via second stage
```
rsahita, line 60: should we call this G-stage?

specification/threatmodel.adoc
```diff
@@ -162,7 +162,7 @@ host software is considered out of scope.
 | Availability | VMM DoS Protection            | Implementation-specific | Not in scope for AP-TEE               | Prevent untrusted code from refusing to run TEE | Not applicable 
 
 | Side Channel | Protected address mapping (controlled side channel)         | Required                 | AP-TEE mode qualifier, cryptography, MMU/MPU, MTT | Similar to memory remapping attacks                                                                 | uSG SIG, AP-TEE to specify 
-| Side Channel | Uarchitectural side channels (branch prediction             | Required                 | uArch state flushing, entropy defenses            | Prevent attacks such as meltdown/spectre (it is difficult to defend agains such attacks in advance) | uSC SIG, AP-TEE specify
+| Side Channel | Architectural side channels (branch prediction             | Required                 | uArch state flushing, entropy defenses            | Prevent attacks such as meltdown/spectre (it is difficult to defend agains such attacks in advance) | uSC SIG, AP-TEE specify
```
rsahita, line 165: this should be Micro-architectural

specification/threatmodel.adoc
rajnesh-kanwal: Both `G-stage` (5 time) and `second stage` (17 times) are used through out the spec. If we change this I would suggest we change all instances of this to `G-stage`. Personally I find `second stage` less confusing. `G-stage` sounds like OS stage translation (1st stage) but then I see the term `VS-stage` only then `G-stage` makes sense.

specification/threatmodel.adoc
rajnesh-kanwal: Fixed.

specification/refarch.adoc
rajnesh-kanwal: Fixed.

specification/threatmodel.adoc
atishp04: The priv spec mentions G-stage and KVM code also revised sometime back to follow the G-stage convention instead of 2nd stage. It would be good to align with those.

rajnesh-kanwal: Sounds good. Let me then move to `G-stage`.

rsahita (approved):

PR #13, opened 2023-03-08
Correct DiceTcbInfo flags description.

jyao1:
Using Uint8 might not be enough, because there might be more than 8 bits defined in the future.
Use bit string to align the definition in DICE specification.

Signed-off-by: Jiewen Yao <jiewen.yao@intel.com>


rsahita (approved):
makes sense - thanks

PR #38, opened 2023-12-19
Fix dynamic register naming and MAX number.

### LINKED ISSUE #36: static/dynamic measurement terminology confusing
The [attestation](https://github.com/riscv-non-isa/riscv-ap-tee/blob/main/specification/attestation.adoc) section defined **static measurements** and **dynamic measurements**, and [sbi_cove](https://github.com/riscv-non-isa/riscv-ap-tee/blob/main/specification/sbi_cove.adoc) section claims that “MeasurementRegisterDescriptor” is “the TCG PCR index this measurement maps to”.

These definition and claim can be easily confused with the TCG [PFP](https://trustedcomputinggroup.org/resource/pc-client-specific-platform-firmware-profile-specification/) and [PTP](https://trustedcomputinggroup.org/resource/pc-client-platform-tpm-profile-ptp-specification/) defined **static PCRs** and **dynamic PCRs**.

People may easily think that COVE **static measurements** means TCG **static PCRs**, and COVE **dynamic measurements** means TCG **dynamic PCRs**.
However, it is NOT the case.

Actually, COVE **dynamic measurements** means TCG **static PCRs**.

I suggest to change the name **static measurements** and **dynamic measurements** to some other terms. (Not use static and dynamic)
For example, 
* COVE **static measurements** => COVE **initial measurements**
* COVE **dynamic measurements** => COVE **runtime measurements**

### LINKED ISSUE #37: MAX_DYNAMIC_MEASUREMENT_REGISTERS  number is too small
According to the definition and https://github.com/riscv-non-isa/riscv-ap-tee/issues/36, the COVE dynamic measurements means TCG static PCRs.

TCG [PFP](https://trustedcomputinggroup.org/resource/pc-client-specific-platform-firmware-profile-specification/) and [PTP](https://trustedcomputinggroup.org/resource/pc-client-platform-tpm-profile-ptp-specification/) defined static PCRs number is from 0 to 16 - totally 17.

However https://github.com/riscv-non-isa/riscv-ap-tee/blob/main/specification/sbi_cove.adoc only define it to be
`#define MAX_DYNAMIC_MEASUREMENT_REGISTERS 8`.

It should match TCG definition and use 17.

jyao1:
Fix #36, static/dynamic naming confusing.
Fix #37, MAX_DYNAMIC_MEASUREMENT_REGISTERS value.


sameo (approved):
Thanks a lot @jyao1

rsahita (approved):
LGTM - thanks @jyao1

PR #41, opened 2024-01-19
specification: Define CoVE function ID namespaces

### LINKED ISSUE #39: CoVE-IO FID namespace
The [CoVE-IO](https://github.com/riscv-non-isa/riscv-ap-tee-io) specification extends the CoVE ABI for supporting device direct assignment into TVMs.

The CoVE-IO ABI extensions require their own SBI FIDs and are currently left TBD in the [spec](https://github.com/riscv-non-isa/riscv-ap-tee-io/blob/main/specification/09-coveio_abi.adoc). I propose that we assign a specific FID range for CoVE-IO, so that the CoVE and CoVE-IO FIDs can be formally defined without overlapping or interleaving.

Since FIDs can fill an entire register (`a6`), one possibility would be to reserve the top half `COVG` and `COVH` EID spaces for CoVE extensions. CoVE-IO could e.g take 6 bits from there and use them as needed.

@rsahita @jyao1 WDYT?

sameo:
Fixes #39

### DIFF
--- a/specification/sbi_cove.adoc
+++ b/specification/sbi_cove.adoc
@@ -9,6 +9,29 @@ extension. This specification introduces three new extensions:
 * CoVE Interrupt Extension (EXT_COVI)
 * CoVE Guest Extension (EXT_COVG)
 
+=== CoVE FID namespaces
+
+External specifications (e.g. the CoVE-IO) may need to extend one of the three
+CoVE SBI extensions with domain specific functions. In order to support that
+requirement each one of the CoVE extensions SBI function IDs (`FID`) range is
+split into separate namespaces.
+
+The main CoVE specification uses FIDs from 0 to 1023 (inclusive), and other
+specifications can extend the CoVE SBI by reserving a FID range after 1024.
+
+Below are the reserved CoVE FID namespaces:
+
+[#table_cove_fid_namespaces]
+.CoVE FID Namespaces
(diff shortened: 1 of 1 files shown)


sameo:
cc @rsahita @jyao1

specification/sbi_cove.adoc
```diff
@@ -9,6 +9,28 @@ extension. This specification introduces three new extensions:
...
+
+[#table_cove_fid_namespaces]
+.CoVE FID Namespaces
+[cols="2,2,6", width=90%, align="center"]
+|===
+2+^s|  SBI Function ID Range                .2+^s| Specification Namespace
+  ^| Start          ^| End
+   | 0 (`0x0`)       | 1023 (`0x3ff`)         | CoVE
+   | 1024 (`0x400`)  | 1087 (`0x43f`)         | CoVE-IO
+|===
```
rsahita, line 32: FIDs over 0x43f may be used for other future extensions right? should add that row and specify that range as unallocated.

rsahita (approved):
LGTM with comments addressed - thanks

specification/sbi_cove.adoc
sameo: Right, good point. This is fixed now.

rsahita (approved):

PR #66, opened 2024-02-27
Updates to address issues 64, 65 (chapter 9)

rsahita:


rsahita:
cc @osman-koyuncu-qualcomm @atishp04

specification/sbi_cove.adoc
```diff
@@ -3,18 +3,40 @@
...
 
-External specifications (e.g. the CoVE-IO) may need to extend one of the three
+Implementations may support more than one confidential supervisor domains. In
+order to support that scenario, as shown in <<cove-fid>>, the most significant 6
+bits of the SBI Function Id for the CoVE Extension may be set by the host to
+specifies the target Supervisor Domain Idenfier/TSM for the CoVH or COVI
+function invoked. For guest-invoked COVG functions, the RDSM must specify the
+Supervisor Domain Identifier for the callee. `a6` bits [16..25] are reserved for
+future use. `a6` bits [0..15] are used by the host to specify the CoVE functions
+as allocated in <<table_cove_fid_namespaces>>.
```
sameo, line 22: Using a register to carry the FID is clean imho, especially from a COVG perspective. But would reserving e.g. a5 for this be preferred to overloading a6 (and somehow not fully following the SBI spec)? I'll let @atishp04 chime in here.

rsahita: I think keeping the SDID with FID seems natural - unless you think 64K space for FID may be limited.

atishp04: One of the guest function (e.g sbi_covg_get_evidence) already uses a5. The SBI calling convention doesn't specifically say something FID convention. So I am okay with overriding it as we don't have to update every function for this. 
At this point, I don't how common this scenario will be where you want to switch supervisor domains at runtime.  Thus, reserving a5 for that may be an overkill.
@avpatel any thoughts about reserving upper bits of FID for SDID ?

specification/sbi_cove.adoc
```diff
@@ -401,6 +425,41 @@ image::tvm_destruction_and_memory_reclamation.svg[]
...
+struct sbiret sbi_cove_get_active_domains(unsigned long active_domains);
+
+-----
+
+Returns a 64 bit vector with bits set for supervisor domains that are active.
+Default value is 1 since supervisor domain 0 is always required (the hosting
+domain). For each non-0 position bit set, the SDID with the value of that bit
+position may be used per the <<cove-fid>> convention to invoke COVH/COVI
+functions. For active domains, the `sbi_covh_get_tsm_info` must be invoked to
+get information about supported TSM capabilities.
```
atishp04, line 448: I think it would be good to describe the responsibilities/expectation when supervisor domain switch happens for a running VM. Probably not here but as a broader section in the spec.

specification/sbi_cove.adoc
```diff
@@ -401,6 +425,41 @@ image::tvm_destruction_and_memory_reclamation.svg[]
 [title= "TVM runtime execution"]
 image::tvm_runtime_execution.svg[]
 
+
+
+== COVE Enumeration Extension (EID #0x434F5645 "COVE")
+
+This common CoVE extension enumerates capabilities for CoVE such as number of
```
atishp04, line 432: Do we really need an extension for this ? This would be queried by the host only. Correct ?

rsahita: right - but I was thinking this extension may serve for other common enumeration functions in the future - such as enumerating capabilities supported before you start using COVH, COVG, COVI and future extensions.

atishp04: All SBI extensions are optional. The more we add, we have to cover the alternate cases in software which makes life bit complicated if there is no use case at this point.

However, if you have an use case discovering broader supervisor domain capabilities apart from domain IDs, we can define an extension for that. Is that a valid assumption that supervisor software may need to invoke this beyond CC scenario ?

ozkoyuncu (approved):

specification/sbi_cove.adoc
rsahita: lines 17-20 do describe the responsibilities of the host and RDSM expected when a function is invoked for/from a specific supervisor domain or workload within a supervisor domain.

specification/sbi_cove.adoc
rsahita: that is correct - supervisor domains may be used for other use cases such as creating a TEE/platform security service partition - perhaps we should call this extension "SUPD" and not specific to "COVE"?

specification/sbi_cove.adoc
rsahita: there are other attestation functions that will also use a5 - so would prefer to not burn a register and use the extra available space in a6 (FID).

specification/sbi_cove.adoc
atishp04: Yeah. That's what I was thinking.

rsahita: For completeness should we describe it here and move it to a separate (sbi 3.0) spec at the time when other extensions are moved?

atishp04: Yes.

specification/sbi_cove.adoc
```diff
@@ -401,6 +425,41 @@ image::tvm_destruction_and_memory_reclamation.svg[]
 [title= "TVM runtime execution"]
 image::tvm_runtime_execution.svg[]
 
+
+
+== COVE Enumeration Extension (EID #0x434F5645 "COVE")
```
rsahita, line 430:
```suggestion
== COVE Enumeration Extension (EID #0x53555044 "SUPD")
```

specification/sbi_cove.adoc
```diff
@@ -401,6 +425,41 @@ image::tvm_destruction_and_memory_reclamation.svg[]
...
+.COVE Get Active Domains
+[cols="2,3", width=90%, align="center", options="header"]
+|===
+| Error code              | Description
+| SBI_SUCCESS             | The operation completed successfully.
+| SBI_ERR_FAILED          | The operation failed for unknown reasons.
+|===
+
+
+
```
rsahita, line 462:
```suggestion

This common extension enumerates capabilities for supervisor domains such as number of
active supervisor domains and capabilities of each supervisor domain, e.g. used for CoVE.

[#sbi_supd_get_active_domains]
=== Function: Enumerate active supervisor domains (FID #0)
[source, C]
-----
struct sbiret sbi_supd_get_active_domains(unsigned long active_domains);

-----

Returns a 64 bit vector with bits set for supervisor domains that are active.
Default value is 1 since supervisor domain 0 is always required (the hosting
domain). For each non-0 position bit set, the SDID with the value of that bit
position may be used per the <<cove-fid>> convention to invoke functions
supported for that domain e.g. COVH. For active domains, other extensions
may be invoked to get capabilities specific to that domain e.g. the
`sbi_covh_get_tsm_info` must be invoked to get information from a supervisor
domain supporting CoVE TSM capabilities.

The possible error codes returned in `sbiret.error` are shown below.

[#table_sbi_supd_get_active_domains]
.SUPD Get Active Domains
[cols="2,3", width=90%, align="center", options="header"]
|===
| Error code              | Description
| SBI_SUCCESS             | The operation completed successfully.
| SBI_ERR_FAILED          | The operation failed for unknown reasons.
|===
```

specification/sbi_cove.adoc
rsahita: @atishp04 ptal

specification/sbi_cove.adoc
```diff
@@ -3,18 +3,40 @@
 [[sbi_tee]]
 == Confidential VM Extension (CoVE) SBI extension proposal
 This section describes the normative Confidential VM Extension(CoVE) SBI
-extension. This specification introduces three new extensions:
+extension. This specification introduces four new extensions:
 
+* CoVE Enumeration Extension (EXT_COVE)
```
atishp04, line 8: nit: EXT_SUPD

atishp04:
Apart from the above nit: lgtm.

specification/sbi_cove.adoc
rsahita, line 8:
```suggestion
* Supervisor Domains Enumeration Extension (EXT_SUPD)
```

rsahita: ack

PR #75, opened 2024-04-05
Broaden the specification's scope to embedded systems with simpler hardware and reduced functionalities

wojciechozga:
Guerney and I introduce in this PR changes to the CoVE spec that make the spec suitable for broader number of use cases, like embedded systems, simpler processors, systems designed under the formal verification regime, such as assured confidential execution (ACE) that we develop as open source project: https://github.com/IBM/ACE-RISCV

- We propose to add a new deployment model to address embedded systems with limited functionality needs. See newly added Appendix D.
- We propose to allow for static and dynamic memory partitioning. Static partitioning can be done with PMP/IOPMP and thus relaxes the requirement for more complex hardware to support CoVE. Consequently, we propose to not mandate MTT.
- We propose to allow for a different (simpler) way of creating a TVM: `Single-step TVM creation`. This would be an alternative to the current ` Multi-step TVM creation`. To make it possible we propose to add a new COVH call: covh_promote_to_tvm()
- We propose to support the local attestation that suits needs of embedded systems that might not have access to network. To make it possible, we propose to add a new COVG call: covg_retrieve_secret()
- We propose to not mandate AIA.


specification/glossary.adoc
```diff
@@ -62,6 +62,8 @@ that hosts multiple mutually distrusting software owned by different tenants.
 
 | MTT | Memory Tracking Table (MTT).
 
+| Relying party | A relying party refer to a system providing access to a secured software application.
```
rsahita, line 65:
```suggestion
| Relying party | An entity that depends on the validity of information about another entity, typically for purposes of authorization cite:[RATS].
```

rsahita: perhaps we should use the RATS definition for attestation terms.

specification/.DS_Store
rsahita, line 1: this file should probably be deleted

specification/appendix_d.adoc
```diff
@@ -0,0 +1,88 @@
...
+[caption="Figure {counter:image}"]
+[title= ": M-mode TSM based deployment model for CoVE"]
+image::img_11.png[align=center]
+
+=== Overview
+<<depd>> shows that the deployment model supports a single confidential supervisor domain in which 
+the TSM runs along with the TSM driver in the M-mode. This single confidential supervisor domain can run multiple 
+TVMs that are isolated from each other using the MMU, i.e., G-stage page tables managed by TSM. TSM isolates the 
+hosting supervisor domain (i.e., OS/VMM and non-confidential applications and VMs) from the confidential supervisor 
+domain (TSM and TVMs) using a hardware memory isolation mechanism, like PMP. The Smmtt extension is therefore not required.
```
rsahita, line 18: Smmtt is not required in this model but not precluded

specification/appendix_d.adoc
```diff
@@ -0,0 +1,88 @@
...
+memory accesses to confidential memory, e.g., PMP. 
+====
+
+=== Static memory partitioning
+The deployment model proposes static partitioning of memory into confidential and non-confidential to simplify 
+formal reasoning about the correctness of the TVM execution and isolation. TSM performs this paritioning early 
+during the boot of the platform, resulting in the following advantages: (1) simplified formal reasoning about the 
+ownership of memory, (2) attestation that covers static system configuration (e.g., values of PMP registers), 
+(3) reduced attack surface between OS/VMM and TSM (narrower ABI). A possible negative consequence of 
+static partitioning is underutilization of resources.
```
rsahita, line 34: perhaps better to clarify - underutilization of memory resources specifically since dynamic page-based conversion of memory between confidential and hosting supervisor domains is not possible in this scenario (unless Smmtt is used).

specification/appendix_d.adoc
```diff
@@ -0,0 +1,88 @@
...
+The TSM must have a list of public keys of those authorized to sign VMs (TVMs) for execution on the platform. The 
+local attestation payload associated with the TVM will be
+signed with a private key. When the VM request to be promoted to a TVM, TSM checks the signature on the LAP. 
+If the signature is not valid, the TSM will not convert the VM and will terminate execution of the
+VM. The method for provisioning these public keys into the TSM is outside the scope of this specification.
+
+====Verifying Platform configuration
+When the creator of the (authorized) TVM does not want it to execute on improperly configured or unauthorized hardware,
+there should be a mechanism supported by hardware (and firmware) for verification. 
+Assuming such hardware exists, for example a trusted platform module (TPM), the VM can be created with an encrypted disk.
```
rsahita, line 68: perhaps better to use a generic term like HW root-of-trust of measurement, storage

specification/attestation.adoc
```diff
@@ -141,7 +145,7 @@ CoVE implementation may look like the following table:
 
      | 5
      | TVM Configuration
-    <| TVM Entry Point and Initial Arguments
+    <| TVM Entry Point, Initial Arguments, Debug CSRs, and Performance CSRs
```
rsahita, line 148: maybe we say - vCPU state?

specification/attestation.adoc
```diff
@@ -65,7 +69,7 @@ The TCB elements for each of them is summarized in the following table:
 .4+.^|Platform
     <| HW RoT for boot, measurement and storage
  .4+<| All M-mode firmwares, including the TSM-driver
-    <| All CPU hardware logic, including MMU and caches
+    <| All CPU hardware logic, including Debug, MMU and caches
```
rsahita, line 72: Maybe we just say All CPU hardware logic  - instead of stating individual sub-systems

specification/overview.adoc
```diff
@@ -18,15 +18,19 @@ of hardware-attested trusted execution environment called TEE Virtual Machines
...
+On processors supporting multiple supervisor domains (the Smmtt extension), the resources
 managed by the hosting supervisor domain (OS/VMM) include memory, CPU, I/O
-resources and platform capabilities to host the TVM workload. The terms
+resources and platform capabilities required to host the TVM workload. The terms
 hosting supervisor domain and OS/VMM are used interchangeably in this
 specification. The underlying memory isolation mechanisms for supervisor domains
-(Smmtt) is agnostic of the number of supervisor domains.
+(Smmtt) is agnostic of the number of supervisor domains. On processors that do not support 
+multiple supervisor domains where Smmtt is not mandated, a single confidential supervisor 
+domain and a single hosting supervisor domain can be supported (see <<depd>>).
```
rsahita, line 33:
```suggestion
domain and a single hosting supervisor domain can be supported (for e.g. <<depd>>, 
other deployment models are also possible).
```

specification/overview.adoc
```diff
@@ -36,25 +40,27 @@ image::img_0.png[]
...
+also called the " *TSM-driver* ", isolates the Confidential Supervisor Domain 
+from all other Supervisor domains and other platform components (non-confidential and
 confidential). The responsibility of the TSM is to enforce the security
-objectives accorded to TEE workloads assigned to that supervisor domain. The
+objectives accorded to TEE workloads (TVMs) assigned to that supervisor domain. The
 VMM is expected to continue to manage the security for non-confidential
 workloads, and importantly the resource-assignment and scheduling management
 functions for all confidential and non-confidential workloads.
+Note that CoVE implementations that support only one Confidential Supervisor
+Domain might combine TSM and TSM-driver functionality in a single deployment unit.
```
rsahita, line 56:
```suggestion
Note that CoVE implementations may partition the TSM and TSM-driver functionality
as required by implementations. For e.g. <<depd>> shows the TSM and TSM-driver
functionality in a single deployment unit.
```

specification/attestation.adoc
wojciechozga: done

specification/appendix_d.adoc
wojciechozga: I added the clarification of underutilization of memory

specification/attestation.adoc
wojciechozga: done

specification/refarch.adoc
```diff
@@ -9,62 +9,70 @@ the properties of the TSM, its instantiation, isolation and operational model
...
+environment (e.g., it does not require Smmtt). <<dep3>> discusses this model in detail. 
 
+=== CoVE Memory Isolation
+Memory isolation for TVMs is orchestrated by the TSM-driver and the TSM in two
+phases: the conversion of memory to confidential memory and the assignment of
+confidential memory (alongwith the enforcement of properties on use) to TVMs.
+To enforce isolation across Host and Confidential supervisor domains, CoVE
+requires isolation of physical memory (that supports paging when enabled). 
+
+CoVE defines two types of memory:
```
rsahita, line 46:
```suggestion
CoVE defines two types of memory:

```

rsahita: need a line to render correctly

specification/refarch.adoc
```diff
@@ -9,62 +9,70 @@ the properties of the TSM, its instantiation, isolation and operational model
...
 including state such as 1st-stage, G-stage page tables.
 * Non-confidential memory - used only for shared data, e.g., communication
 between the TVM/TSM and the non-TCB host software and/or non-TCB IO devices.
 
-The TSM COVH ABI provides interfaces to the OS/VMM to convert / donate
-memory from the hosting supervisor domain to the confidential supervisor domain.
+The split of memory into confidential and non-confidential may be static or dynamic. 
+Static partitioning occurs during platform initialization while the dynamic partitioning 
+occurs during runtime. The dynamic partitioning allows for better resource utilization 
+at the cost of more complex hardware and software implementation.
```
rsahita, line 55:
```suggestion
occurs during runtime. The dynamic partitioning allows for better resource utilization, 
but has additional hardware and software requirements. 
```

specification/sbi_cove.adoc
```diff
@@ -109,6 +122,15 @@ The host is responsible for the following memory management functions:
 . Mapping zero-page confidential pages to the TVM regions.
 . Mapping non-confidential pages TVM-defined regions for shared-pages / MMIO.
 
+[NOTE]
+====
+The division into confidential and non-confidential memory might be done statically or dynamically. 
+CoVE implementations that support only static partitioning of confidential and 
+non-confidential memory (for example, high-assurance systems that must be formally verified, see <<appendix_d>>)
```
rsahita, line 129:
```suggestion
non-confidential memory (for example, see <<appendix_d>>) 
```

rsahita: systems with dynamic confidential memory may also be formally verified and high assurance so I removed that qualification from this example reference.

specification/sbi_cove.adoc
```diff
@@ -1651,6 +1677,65 @@ COVE-Guest calls cause a trap to the TSM. TSM should do any processing required
...
+| SBI_ERR_FAILED          | The operation failed for unknown reasons.
+|===
+
+[#covg_retrieve_secret]
+=== Function: COVE Guest Retrieve Secret (FID #0)
+[source, C]
+-----
+struct sbiret covg_retrieve_secret();
+-----
+Requests a TSM a secret available after successful local attestation. TSM reads this secret during
```
rsahita, line 1722:
```suggestion
Requests TSM for a secret available after successful local attestation. TSM reads this secret during
```

specification/swlifecycle.adoc
```diff
@@ -8,13 +8,17 @@ including the OS/VMM interactions with the TSM.
...
 be able to use all of the system memory as confidential memory, as long as
 the platform access-control mechanisms are applicable to all the available
-memory on the system. The TSM allows the OS/VMM to manage the assignment of
-confidential memory by providing a two stage TEE memory management model:
+memory on the system. In CoVE implementations that support dynamic conversion 
+between confidential and non-confidential memory, the TSM might allow the 
+OS/VMM to manage the assignment of confidential memory by providing a two stage 
+TEE memory management model, described below. However, this memory management 
+model is not required for CoVE implementations that statically partition memory into
+confidential and non-confidential, e.g., durich platform initialization.
```
rsahita, line 21:
```suggestion
confidential and non-confidential, e.g., during platform initialization.
```

rsahita (changes requested):
With some of the minor review comments addressed, LGTM

specification/sbi_cove.adoc
```diff
@@ -1651,6 +1677,65 @@ COVE-Guest calls cause a trap to the TSM. TSM should do any processing required
 and then must forward the ECALL to the host with `scause` set to ECALL, `a7` set
 to EID, `a6` set to FID (See <<cove-fid>>), `a0`-`a5` set to ECALL args.
 
+[#sbi_covg_promote_to_tvm]
+=== Function: COVE Guest Promote to TVM (FID #0)
+[source, C]
+-----
+struct sbiret sbi_covg_promote_to_tvm(unsigned long fdt_addr,
+                                      unsigned long lap_addr);
+-----
```
eckhard-delfs-qualcomm, line 1686: Is this function intended to be applicable only in the context of Appendix D?
If planned as a general extension, then the extended scope of the COVEG-ABI needs to be reflected in Figure 1 (similar as depicted img_11). Alternatively, it may be worth considering defining a dedicated ABI for host VM originated COVE calls.

specification/sbi_cove.adoc
```diff
@@ -1651,6 +1677,65 @@ COVE-Guest calls cause a trap to the TSM. TSM should do any processing required
...
+| Error code              | Description
+| SBI_SUCCESS             | The operation completed successfully.
+| SBI_ERR_INVALID_ADDRESS | `fdt_addr` was invalid.
+| SBI_ERR_AUTH            | Local attestation failed.
+| SBI_ERR_OUT_OF_MEMORY   | Not enough memory to store TVM
+| SBI_ERR_FAILED          | The operation failed for unknown reasons.
+|===
+
+[#covg_retrieve_secret]
+=== Function: COVE Guest Retrieve Secret (FID #0)
```
eckhard-delfs-qualcomm, line 1717: FID #1?

specification/sbi_cove.adoc
```diff
@@ -94,10 +93,24 @@ the current status of the TSM. The TSM must be in `TSM_READY` in order to
...
-set to `TVM_INITIALIZING`.
-The host must assign confidential memory for page tables, payload mapping, and
-vCPUs before it can be transitioned into the `TVM_RUNNABLE` state.
+A TVM might be created in two ways: (1) a non-confidential VM requests TSM to be promoted 
+to a TVM, (2) VMM assembles a TVM under the TSM supervision.
+
+The promotion of a non-confidential VM to a TVM happens when a regular (non-confidential) VM
+executes a `sbi_covg_promote_to_tvm()` call. TSM handles this call by copying the entire
+VMs state (the CPU state, VM memory, page tables) into confidential memory. This is performed as a single 
+operation after which the newly created TSM is in the `TVM_RUNNABLE` state.
```
eckhard-delfs-qualcomm, line 102: newly created TVM

specification/sbi_cove.adoc
```diff
@@ -94,10 +93,24 @@ the current status of the TSM. The TSM must be in `TSM_READY` in order to
...
+VMs state (the CPU state, VM memory, page tables) into confidential memory. This is performed as a single 
+operation after which the newly created TSM is in the `TVM_RUNNABLE` state. 
+
+[NOTE]
+====
+Creating a TVM in a single step reduces the number of ABI calls, thus simplifies the TSM and OS/VMM implementation. 
+However, it will block the CPU for the duration of the TVM creation.
+====
+
+In the second approach, VMM assembles the TVM in a sequences of calls to TSM. The first
```
eckhard-delfs-qualcomm, line 110: in a sequence

specification/appendix_d.adoc
```diff
@@ -0,0 +1,86 @@
...
+local attestation payload associated with the TVM will be
+signed with a private key. When the VM request to be promoted to a TVM, TSM checks the signature on the LAP. 
+If the signature is not valid, the TSM will not convert the VM and will terminate execution of the
+VM. The method for provisioning these public keys into the TSM is outside the scope of this specification.
+
+==== Verifying Platform configuration
+When the creator of the (authorized) TVM does not want it to execute on improperly configured or unauthorized hardware, there should be a mechanism supported by hardware (and firmware) for verification. 
+Assuming presence of the hardware root-of-trust for measurement and hardware root-of-trust for storage, the VM can be created with an encrypted disk and the key that is used to decrypt the disk can be sealed to the measurments of the platform. 
+The creator of the VM using the specifications for the platform decides what values are required in order for the key to be released. 
+When the request to promote VM to a TVM is called, the TSM useals the key with help of the hardware root-of-trust.
```
eckhard-delfs-qualcomm, line 72: unseals

specification/appendix_d.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+==== Verifying Platform configuration
+When the creator of the (authorized) TVM does not want it to execute on improperly configured or unauthorized hardware, there should be a mechanism supported by hardware (and firmware) for verification. 
+Assuming presence of the hardware root-of-trust for measurement and hardware root-of-trust for storage, the VM can be created with an encrypted disk and the key that is used to decrypt the disk can be sealed to the measurments of the platform. 
+The creator of the VM using the specifications for the platform decides what values are required in order for the key to be released. 
+When the request to promote VM to a TVM is called, the TSM useals the key with help of the hardware root-of-trust. 
+If the request fails, the promotion of VM to TVM fails and the TSM returns error to the VM. 
+When the request succeeds, the TSM completes the transistion by informing the OS/VMM about the VM promotion to TVM, and the control eventually returns to the TVM. At the point where the TVM needs to decrypt its disk (e.g., for mounting the filesystem), the TVM utilizes an ABI call (`covg_retrieve_secret()`) to TSM to retrieve the decryption key.
```
eckhard-delfs-qualcomm, line 74: typo: the TSM completes the transition...
also: required sbi call to inform VMM should be listed here.

specification/sbi_cove.adoc
wojciechozga: fixed

specification/sbi_cove.adoc
wojciechozga: fixed, thank you.

rsahita (approved):

specification/sbi_cove.adoc
gdhh: This is not the case, it could be use by others

gdhh: Figure 1, figure 2, and the deployment model in our figure are specific to deployment models. IF the question is whether deployment models 1 & 2 could use a one-step conversion process, the answer is yes. If for that reason you want to show the call, the I am ok with showing it in Figure 1. But some how indicating it is optional.

specification/appendix_d.adoc
gdhh: The flow should be VM request promotion, TSM promotes, TSM reflect promotion call to hypervisor. No new call is required. We will make this clearer.

specification/sbi_cove.adoc
wojciechozga: I adjusted FIDs to be unique in COVG after the addition of new calls.

specification/appendix_d.adoc
wojciechozga: I fixed the typo and adjusted the description of the call in the `COVE SBI extensions` chapter.

rsahita (approved):

atishp04:
Few broader questions:

1. As per the discussion in the RVI meeting: Keeping the TSM in M-mode helps in formal verification. As per my understanding it requires OpenSBI to be formally verified as well. If we run OpenSBI in M-mode and TSM in HS mode, the formal verification effort will be the same. Isn't it ?

Do we require more involved formal verification because of privilege mode changes ? 

Note: I don't have formal verification background. Thus, I may be asking stupid questions :)

specification/overview.adoc
```diff
@@ -28,7 +28,8 @@ managed by the hosting supervisor domain (OS/VMM) include memory, CPU, I/O
 resources and platform capabilities required to host the TVM workload. The terms
 hosting supervisor domain and OS/VMM are used interchangeably in this
 specification. The underlying memory isolation mechanisms for supervisor domains
-(Smmtt) is agnostic of the number of supervisor domains. On processors that 
+(Smmtt) is agnostic of the number of supervisor domains. On processors with 
+hypervisor mode that
```
atishp04: nit: hypervisor(h) extension ?

specification/swlifecycle.adoc
```diff
@@ -25,13 +25,24 @@ controls described earlier).
...
+
+In deployment models 1 and 2, the host OS/VMM may create a new TVM by 
+allocating and initializing a TVM
 using the `sbi_covh_create_tvm()` function. An initial set of memory pages are
 granted to the TSM and tracked as TEE pages associated with that TVM from
 that point onwards until the TVM is destroyed via the `sbi_covh_destroy_tvm()`
-function.
+function. In deployment model 3, all TVMs start execution as a VM.
+To become a TVM, the VM request that the TSM change it into
+a TVM using the sbi_coveh_promote_TVM call.
```
atishp04: Can the promotion be anytime during the run of VM or any certain points ?

specification/swlifecycle.adoc
```diff
@@ -629,6 +656,15 @@ Confidential-Unassigned state.
 and returns the PA as an Non-Conf page to the VMM.
 * VMM translations to the PA (via 1st or G stage mappings) may be created now.
 
+[Note]
+====
+For deployment model 3 when no Smmtt is available memory is dynamically partitions. TVM shutdown is
+simler, the VMM can cause all pages to be reclaimed via the sbi_covh_reclaim_pages which will
```
atishp04: nit: simler/simpler ?

wojciechozga: Hi Atish, you are reviewing the outdated version of the spec. These typos have been fixed.

specification/swlifecycle.adoc
atishp04: The section in the appendix describes this. 
"To reduce the complexity of the TSM implementation, TSM creates TVM as a result of a single operation triggered with the `sbi_covg_promote_to_tvm()` call. Specifically, OS/VMM initializes and starts a regular VM, which early during the boot process requests to be promoted to TVM. This call is reflected to TSM that copies TVM data, the page table configuration, and the vCPU state into confidential memory."

wojciechozga:
> Few broader questions:
> 
> 1. As per the discussion in the RVI meeting: Keeping the TSM in M-mode helps in formal verification. As per my understanding it requires OpenSBI to be formally verified as well. If we run OpenSBI in M-mode and TSM in HS mode, the formal verification effort will be the same. Isn't it ?
> 
> Do we require more involved formal verification because of privilege mode changes ?
> 
> If deployment model 3 is aimed for embedded use case, TSM can just run HS mode without any M-mode involvement. This also reduces the TCB to be verified.
> 
> Note: I don't have formal verification background. Thus, I may be asking stupid questions :)

Keeping TSM in M-mode helps with formal verification because:  (1) it simplifies reasoning because the security domains run in different privileged modes, (2) there are no context switches/switches between the M-mode and HS-mode that will be non-trivial to reason about, because one would have to reason about the ISA machine state for that transition, (3) it does not require to reason about resource ownership split between TSM running in HS-mode and a more privileged M-mode software  (4) it reduces the number of TSM’s source line of code (SLoC) written in assembly; assembly is more difficult for formal reasoning.

You might run TSM in HS-mode in embedded scenario and try to formally verify some of its property. The deployment model 3 in Appendix D shows a variant of CoVE where TSM runs in M-mode because it is more practical for formally reasoning, as stated above. 

The decision whether to verify OpenSBI or not depends on the goals of using formal methods for CoVE. These can vary. For example, for strong security claims you might have to prove all M-mode software but probably still have to axiomatize the hardware implementation. On the other hand, one might try to use formal methods for verifying functional correctness of specific parts of the implementation, so verifying only a subset of TSM functionality would be ok. Different certification levels (FIPS, Common Criteria) might have different requirements for what is expected to be proven.

specification/swlifecycle.adoc
```diff
@@ -65,6 +76,13 @@ measurement of the TVM via `sbi_covh_finalize_tvm()`. The TSM prevents any
 TVM virtual harts from being entered until the TVM initialization is
 finalized.
 
+=== Deployment Model 3
+
+When a TVM context is created and intialized by using the sbi_coveh_promote_TVM
+call all of the pages of the TVM are allocate at creation. The default
```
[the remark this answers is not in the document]
wojciechozga: Thanks for reviewing. The typo was fixed in one of the previous commits, but the review was made on an outdated version of the PR.

specification/swlifecycle.adoc
wojciechozga: Thanks for reviewing. I updated the description of the time when the promotion should take place in the section `Single-step TVM Creation` in `TVM Lifecycle` chapter.

specification/overview.adoc
wojciechozga: Thanks for reviewing. This text was already rewritten in one of the previous commits. Could you please have a look to the newest version?

gdhh:
> Few broader questions:
> 
> 1. As per the discussion in the RVI meeting: Keeping the TSM in M-mode helps in formal verification. As per my understanding it requires OpenSBI to be formally verified as well. If we run OpenSBI in M-mode and TSM in HS mode, the formal verification effort will be the same. Isn't it ?
> 
> Do we require more involved formal verification because of privilege mode changes ?
> 
> If deployment model 3 is aimed for embedded use case, TSM can just run HS mode without any M-mode involvement. This also reduces the TCB to be verified.
> 
> Note: I don't have formal verification background. Thus, I may be asking stupid questions :)

Wojciech also replied to this. Fundamentally it appears that are proposing another deployment model. On a processor without smmtt running the TSM in HS mode implies either that it is running beside the hypervisor or that it has replaced the hypervisor. Either approach would mean more code in the TSM and therefor more complexity to verify. This approach could be simplified by moving the Hypervisor to S mode (a form of deprivilege). This would provide hardware separation between the TSM and the hypervisor however, all request from all VMs would have to be routed via the TSM. Whereas in the current approach only request from confidential VM are routed through the TSM.  If this is a correct understanding of your proposal I agree that if we are successful in proving the TSM in our proposed model that proof could be leveraged for the model you have proposed. But I do not think that the current work we are doing would be sufficient to prove the new model. I believe that there would be more work, even though OpenSBI is running separately in M mode.

specification/overview.adoc
atishp04: My bad. I see it now. I am not sure why github was showing me the older commits.

rsahita (approved):

specification/sbi_cove.adoc
rsahita: I think its fine to keep it as part of COVG - I think it is understood (but can be added in the text) that this call is expected to originate from a VM not operating as a TVM. if invoked from a TVM this ABI will return an error.

wojciechozga: I modified the description of this call to make it clear that this call is intended for a VM and not a TVM.

eckhard-delfs-qualcomm: Thanks @gdhh, @wojciechozga, @rsahita for the feedback. 
Still, looking at chapter 12, I believe it may cause confusion to readers. This chapter introduces the scope of COVG by saying that it allows TVMs to communicate with their TSM. But the first function (sbi_covg_promote_to_tvm) in the following section does not comply and even states that it must not be called by a TVM. IMHO it would help if this function was pushed to the end of the chapter and introduced with some clarifying words regarding the intended deployment model (with reference to appendix D) and confirmation of applicability to the primary deployment model in Figure 1.

gdhh:
> 
> Not necessarily if the non-confidential VMs are run as a nested VM. But TSM does have to some additional H-mode state to enable that.
> 
> > request from confidential VM are routed through the TSM. (supporting confidential computation has minimal impact on normal VMs.) If this is a correct understanding of your proposal I agree that if we are successful in proving the TSM in our proposed model that proof could be leveraged for the model you have proposed. But I do not think that the current work we are doing would be sufficient to prove the new model. I believe that there would be more work, even though OpenSBI is running separately in M mode.
> 
> Agreed. I was just trying to understand the challenges of formal verification between running different models. It seems it boils down to more complex TSM code/additional privilege state management vs more code in M-mode.
> 

I think this is a reasonable summary.

> As per my understanding from your and @wojciechozga's response, the later is easier. But if we can axiomatize hardware implementation and verifying only a subset of TSM functionality is okay, it shouldn't matter. Isn't it ?

I believe that by verifying only a subset of TSM functionally you mean that only the functionality that affects TVM. I do not think that this is a good approach.  Because the VMs and TVMs would be sharing the same TSM the side channel issue discussed below has to be considered.

If by nested VMs you mean running a second level hypervisor that services all non confidential VMs. That could possibly work (assuming the correct hardware support). However, it is more complex than what we have proposed. The other issue is that unless we change the hardware implementation, the second level hypervisor would possibly have to get services from the TSM.  

In addition to proof complexity, once the TSM is servicing anyone other than TVMs we open up the possibility of side channels. In particular we would have to show that there is no way to leak information between the TVS and the VMs (or second level hypervisor). Essentially, you need to show that nothing the VMs do will influence the TVMs and that none of the VM request will be influenced by the TVMs. I believe that it is possible to build a system with this approach where this is true. But it puts an additional architecture and proof burden on the system that the approach we have proposed does not have.  The TSM only services one call from the VMs. That call results in the VM either being terminated because it fails attestation or becoming a TVM. Properly chosen cryptographic algorithms and key length (and key material) means that try fail semantics to break into the system would take too long to be practical.

Our current proof approach is not without side channel issues. We have to show that TVMs do not leak information to one another because they share the same TSM. Our approach to this might also work for an approach where the TSM services both, but that is a complexity we are avoiding. My statement on the required additional proof work is only an observation of the amount of work to be done and not a judgement as to whether it can be done.  We are still in the process of proving the current approach so we do not know whether we will be successful, but we suspect so. We want to add some additional capabilities but not until we have a proven base.

specification/sbi_cove.adoc
```diff
@@ -1686,20 +1686,20 @@ to EID, `a6` set to FID (See <<cove-fid>>), `a0`-`a5` set to ECALL args.
...
+is known and the resulting integrity measurement hash is deterministic. The TSM uses the VM page table configuration
+to copy VM's data to confidential memory. TSM extracts from FDT information about the number of vCPUs, allocating
 required amount of memory and populating one with the vCPU state of the caller.
 
-On success, TSM reflects an SBI call `sbi_register_tvm(unsigned long tvm_guest_id)` to the hypervisor.
-The hypervisor marks the VM as a TVM and frees the VM memory. For now on, OS/VMM interacts with a TVM
-with help of the TSM using the COVH ABI. It resumes TVM using the `sbi_covh_run_tvm_vcpu()`
-call.
+On success, the TSM reflects an SBI call `sbi_register_tvm(unsigned long tvm_guest_id)` to the OS/VMM, which marks
+the VM as a TVM and frees the VM memory. For now on, OS/VMM interacts with this TVM via the TSM using the
```
gdhh, line 1705: Line 1701 should be : 
"the VM as a TVM and frees the VM memory. From now on, OS/VMM interacts with this TVM via the TSM using the". from not for

specification/sbi_cove.adoc
wojciechozga: @eckhard-delfs-qualcomm, thank your for your input.
Following your suggestions, I moved the `sbi_covg_promote_to_tvm()` to the end of the COVG ABI section and modified its description to make it clear that (1) it is intended for deployment model 3, (2) it is applicable (but optional) to deployment model 1 as well, (3) it is intended for VMs not TVMs.

specification/sbi_cove.adoc
wojciechozga: fixed

rsahita (approved):

specification/sbi_cove.adoc
```diff
@@ -2077,6 +2137,47 @@ register indexes.
...
+call, otherwise this ABI will return error. It is primarily inteded for COVE deployment models that require 
+single-step TVM creation (e.g., see <<dep3>>). Deployment models that offer multi-step TVM creation 
+(e.g., <<dep1>>) may, but are not required to, support this ABI as an additional mechanism for creating TVMs. 
+
+The `fdt_addr` is the 8-bytes aligned guest physical address of the flattened device tree (FDT). `tap_addr` 
+is the 8-bytes aligned guest physical address of the `TVM attestation payload` used for local attestation. 
+VMs that do not require local attestation (only the remote attestation) must write `0` to `tap_addr`.
+
+The VM must make this call early in the boot process so that the contents of the memory are known and the 
+resulting integrity measurement hashes are deterministic. The TSM receives this call and uses the VM page
```
rsahita, line 2157: this sentence is odd - the contents of memory are known to whom? suggest re-state as:

"The VM should issue this call as early in the boot process as possible to minimize changes in contents of memory so that the resulting integrity measurement hashes are deterministic."

specification/sbi_cove.adoc
```diff
@@ -2077,6 +2137,47 @@ register indexes.
...
+VMs that do not require local attestation (only the remote attestation) must write `0` to `tap_addr`.
+
+The VM must make this call early in the boot process so that the contents of the memory are known and the 
+resulting integrity measurement hashes are deterministic. The TSM receives this call and uses the VM page 
+table configuration to copy VM's data to confidential memory. TSM extracts information about the number of 
+vCPUs from FDT, allocates required amount of memory, and populates one of the vCPUs state with the state of 
+the caller's thread.
+
+If successful, the TSM reflects an SBI call `sbi_register_tvm(unsigned long tvm_guest_id)` to the OS/VMM, 
+which marks the VM as TVM and frees the VM memory. As a result of this call the OS/VMM will interact with this TVM via
```
rsahita, line 2163: how does the TSM reflect the status change of the VM to the OS/VMM? due to this flow, I was wondering if the promote call should be bridged/proxied by the OS/VMM (since it is untrusted anyway).

rsahita (commented):
after thinking some more, had a couple questions on the promote abi

specification/sbi_cove.adoc
wojciechozga: I modified the text according to your suggestion.

specification/sbi_cove.adoc
gdhh: The TSM uses SBI_register_tvm call to notify the OS/VMM that the VM is now a TSM. Since this appears to the hypervisor as a call from the VM (we reflect it as a hyper call), it is possible that a normal VM could make this call. In which case it will result in  termination because the hypervisor will free all the VMs memory as a result of this call. When the hypervisor tries to resume tvm_guest_id, it will get an error because the tvm_guest_id most likely does not exist.

gdhh: Issuing the call via the hypervisor, as you suggested, is another approach but requires changes to the OS/VMM (Linux/KVM) not required by the present approach. Another possible approach, which we have not taken, is to introduce a new instruction for calls directly to the TSM, instead we have leveraged the environmental calls.

wojciechozga: Proxying the call through the OS/VMM is an option, but it would require some boilerplate code in OS/VMM (KVM) implementation. Let me prototype what KVM changes this approach would require.

wojciechozga: @rsahita  @eckhard-delfs-qualcomm 
We discussed internally to proxy the promote_to_tvm() call over the OS/VMM and implemented a PoC to see what is the amount of code that we would require to place in the OS/VMM (KVM) to reflect this call. We believe it is reasonable and the . We will work on the COVE spec to define promote_to_tvm() in the COVH SBI and remove it from the COVG SBI. We will have it done tomorrow (Friday, April 19).

specification/sbi_cove.adoc
```diff
@@ -843,8 +851,59 @@ The possible error codes returned in `sbiret.error` are shown below.
...
+-----
+struct sbiret sbi_covh_promote_to_tvm(unsigned long fdt_addr,
+                                      unsigned long number_of_vcpus,
+                                      unsigned long tap_addr);
+-----
+This intrisic is used by the host to promote a VM to a TVM. It is primarily inteded for CoVE deployment 
+models that require the single-step TVM creation (e.g., see <<dep3>>). Deployment models that offer the multi-step TVM creation 
+(e.g., <<dep1>>) may, but are not required to, support this ABI as an additional mechanism for creating TVMs. 
+
+The `fdt_addr` is the 8-bytes aligned guest physical address of the flattened device tree (FDT). The `fdt_addr` is
```
rsahita, line 866: typo for param 2 
also a q - does not the fdt describe the number of vcpus also? what if that and the param 2 do not agree?

rsahita (changes requested):
lgtm = w/ one comment - thanks

gdhh (commented):
line 198 should be: The NACL memory interface is between the TSM and the host, and the TSM is

gdhh (approved):

specification/sbi_cove.adoc
wojciechozga: Passing the number of vcpus as a parameter could simplify the TSM implementation, but indeed might lead to consistency issues (mismatch with the content of FDT). Thus, I removed this parameter.

specification/sbi_cove.adoc
```diff
@@ -97,10 +97,10 @@ A TVM might be created in two ways:
 (1) the OS/VMM requests the TSM to promote an existing VM to a TVM using a single COVH call, 
 (2) the OS/VMM assembles a TVM under the TSM's supervision in a sequence of COVH calls. 
 
-To promote a non-confidential VM to a TVM, the OS/VMM invokes the `sbi_covh_promote_to_tvm()` call,
-presenting the VM state (the CPU state, VM's data, page tables) to the TSM. The TSM handles this call 
-by copying the entire VMs state into confidential memory. This is performed as a single operation after 
-which the newly created TVM is in the `TVM_RUNNABLE` state. 
+To promote a non-confidential VM to a TVM, the OS/VMM invokes the `sbi_covh_promote_to_tvm()`,
+presenting the VM's state (the CPU state, VM's data, page tables) to the TSM. The TSM then 
+copies the entire VM's state into confidential memory. This is performed as a single operation after
```
rsahita, line 102:
```suggestion
measures and then copies the entire VM's state into confidential memory. This is performed as a single operation after 
```

rsahita (approved):

rsahita (approved):
seems ok to merge now

PR #76, opened 2024-04-24
Cleanup to create RC for ARC review

rsahita:
Folded asciidoc text to 80 char width
Replaced outdated sequence diagrams with plantuml
Fixed references etc.


sameo (approved):
Nice cleanup. I really like the kroki integration, we'll switch to that for CoVE-IO.

PR #82, opened 2024-07-16
TSM capabilities and minor editorial changes

wojciechozga:
We propose to extend the COVH get_tsm_info() call with additional information called `TSM capabilities` to simplify implementation of a hypervisor that supports different CoVE deployment models. Specifically, TSM informs the hypervisor via the `tsm_info` structure what capabilities is supports, such as support for single- or mult-step TVM creation, local and/or remote attestation, legacy interrupt handling or AIA, static or dynamic memory allocation.

Additionally, we provide a set of minor editorial changes that clarify: (1) how interrupts are injected if AIA is not supported, (2) what properties local attestation can offer for embedded platforms, (3) how TVM enables or disables injection of all interrupts (wildcard), (4) how MMIO regions are specified, (5) how TVM retrieves local attestation secret.

### DIFF
--- a/src/sbi_cove.adoc
+++ b/src/sbi_cove.adoc
@@ -2355,7 +2355,8 @@ confidential memory.
 === Function: COVE Guest Retrieve Secret (FID #9)
 [source, C]
 -----
-struct sbiret covg_retrieve_secret();
+struct sbiret covg_retrieve_secret(unsigned long buf_addr,
+                                   unsigned long buf_size);
 -----
 Requests TSM for a secret available after successful local attestation. TSM
 reads this secret during local attestation from the `TVM attestation payload`
@@ -2363,6 +2364,11 @@ reads this secret during local attestation from the `TVM attestation payload`
 creation via `sbi_covh_promote_to_tvm()`. Only the TVMs that were correctly
 authenticated and authorized during local attestation can receive the secret
 embedded in TAP.
+
+`buf_addr` is the 4KiB-aligned address in the TVM address space to which the 
+TSM will write the secret.
+`buf_size` is the length of the buffer. The buffer cannot be larger than the 
+4KiB page size.
 
 This ABI will become part of the `Sealing Interface` planned for the CoVE in
 version 2.0.
@@ -2376,6 +2382,8 @@ The possible error codes are shown below.
 |===
 | Error code              | Description
 | SBI_SUCCESS             | The operation completed successfully.
+| SBI_ERR_INVALID_ADDRESS | `buf_addr_out` is not a valid guest physical address or is not correctly aligned.
+| SBI_ERR_INVALID_PARAM   | The buffer's size is too small to store the secret or larger than the 4KiB page.
 | SBI_ERR_AUTH            | Local attestation failed.
 |===
 
@@ -2387,7 +2395,7 @@ struct sbiret sbi_covg_read_measurememt(unsigned long msmt_buf_addr_out,
                                              unsigned long msmt_buf_size,
                                              unsigned long msmt_index);
 -------
-This intrisic returns a the TVM measurement register value for the `msmt_index`
+This intrisic returns a TVM measurement register value for the `msmt_index`
 measurement register. TVMs can read both initial and runtime measurement
 register values back.
 
@@ -2704,6 +2712,10 @@ intrinsic is used by a TVM to get an attestation evidence to
 report to a remote relying party. It returns an attestation certificate signed
 by the TSM attestation key, and includes the TVM attestation evidence. The TSM
 attestion key is also included in the reported TSM token.
+
+| <<covg_retrieve_secret, covg_retrieve_secret>>      | TVM reads a secret 
+available after successful local attestation. TSM writes the secret to the
+buffer specified by the TVM.
 
 | <<sbi_covg_read_measurement, sbi_covg_read_measurement>>      | This
 intrisic returns a the TVM measurement register value for the `msmt_index`
--- a/src/sbi_cove.adoc
+++ b/src/sbi_cove.adoc
@@ -775,7 +775,8 @@ enum tsm_state {
 #define COVE_TSM_CAP_ATTESTATION_LOCAL   0x1
 #define COVE_TSM_CAP_ATTESTATION_REMOTE  0x2
 #define COVE_TSM_CAP_AIA                 0x3
-#define COVE_TSM_CAP_MEMORY_ALLOCATION   0x4  
+#define COVE_TSM_CAP_MRIF                0x4
+#define COVE_TSM_CAP_MEMORY_ALLOCATION   0x5  
 
 /* Reserved IDs for TSM implementations */
 #define COVE_TSM_IMPL_SALUS            0x1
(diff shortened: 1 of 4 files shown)


src/sbi_cove.adoc
```diff
@@ -770,10 +770,16 @@ enum tsm_state {
...
 
-#define COVE_TSM_CAP_PROMOTE_TO_TVM       0x1
-#define COVE_TSM_CAP_LOCAL_ATTESTATION    0x2
-#define COVE_TSM_CAP_AIA                  0x3
-#define COVE_TSM_CAP_DYN_PAGE_CONVERSION  0x4
+/* Below specifies bit position in the `tsm_capabilities` field */
+#define COVE_TSM_CAP_PROMOTE_TVM         0x0
+#define COVE_TSM_CAP_ATTESTATION_LOCAL   0x1
+#define COVE_TSM_CAP_ATTESTATION_REMOTE  0x2
+#define COVE_TSM_CAP_AIA                 0x3
```
gdhh, line 777: In the review today we discovered that a capability is missing. We need to add:
#define COVE_TSM_MRIF 0x4
and chang the memory allocation to 0x5

src/sbi_cove.adoc
```diff
@@ -824,11 +830,18 @@ struct tsm_info {
...
-| 4   | COVE_TSM_CAP_DYN_PAGE_CONVERSION   | TSM supports dynamic page conversion.
+| Bit   | Name                               | Value              | Definition
+| 0     | COVE_TSM_CAP_PROMOTE_TVM           | `0`                | TSM supports multi-step TVM creation.
+| 0     | COVE_TSM_CAP_PROMOTE_TVM           | `1`                | TSM supports single-step TVM creation.
+| 1     | COVE_TSM_CAP_ATTESTATION_LOCAL     | `0`                | TSM does not support local attestation.
+| 1     | COVE_TSM_CAP_ATTESTATION_LOCAL     | `1`                | TSM supports local attestation.
+| 2     | COVE_TSM_CAP_ATTESTATION_REMOTE    | `0`                | TSM does not support remote attestation.
+| 2     | COVE_TSM_CAP_ATTESTATION_REMOTE    | `1`                | TSM supports remote attestation.
+| 3     | COVE_TSM_CAP_AIA                   | `0`                | TSM supports legacy interrupt handling.
+| 3     | COVE_TSM_CAP_AIA                   | `1`                | TSM supports RISC-V AIA.
```
gdhh, line 841: Add two lines here 
| 4 | COVE_TSM_CAP_MRFI.       | '1'  | TSM supports RISC-V AIA with MRFI
| 4 | COVE_TSM_CAP_MRFI.       | '0' | TSM supports RISC-V AIA without MRFI

putting these lines here changes the next bit to 5 and the reserved to 6 ...63

src/sbi_cove.adoc
```diff
@@ -770,10 +770,16 @@ enum tsm_state {
...
+/* Below specifies bit position in the `tsm_capabilities` field */
+#define COVE_TSM_CAP_PROMOTE_TVM         0x0
+#define COVE_TSM_CAP_ATTESTATION_LOCAL   0x1
+#define COVE_TSM_CAP_ATTESTATION_REMOTE  0x2
+#define COVE_TSM_CAP_AIA                 0x3
+#define COVE_TSM_CAP_MEMORY_ALLOCATION   0x4  
+
+/* Reserved IDs for TSM implementations */
+#define COVE_TSM_IMPL_SALUS            0x1
+#define COVE_TSM_IMPL_ACE              0x2
```
gdhh, line 782: Add a table to cover TSM implementation. It will probably have the following columns:
| Name       | Value.     |  Definition           | 
| COVE_TSM_IMPL_SALUS | 1 | Salus COVE implementation
|COVE_TSM_IMPL_ACE.     | 2 |. ACE COVE implementation
|  Reserved |  | Future COVE implementations should be assigned in this space.

or something similar, it should be consistent with other tables in the specification

src/sbi_cove.adoc
wojciechozga: I modified it accordingly.

src/sbi_cove.adoc
wojciechozga: I modified it accordingly.

src/sbi_cove.adoc
wojciechozga: I modified it accordingly.

rsahita (approved):
Lgtm - thanks for the updates from the 7-16 meeting review

src/sbi_cove.adoc
```diff
@@ -761,30 +769,66 @@ enum tsm_state {
...
+    uint64_t tvm_max_vcpus;
     /*
-     * The number of 4KB pages which must be donated to the TSM when
-     * creating a new vCPU.
+     * The number of 4KiB pages which must be donated to the TSM when creating
+     * a new vCPU. `0` if the TSM does not support the dynamic page conversion
+     * capability.
      */
-    unsigned long tvm_vcpu_state_pages;
+    uint64_t tvm_vcpu_state_pages;
```
atishp04, line 821: Any specific reasons to change these to uint64_t ? In the SBI spec, we try to use unsigned long if other data type is not warranted. This allows the spec to easier to implement on RV32 (probably not applicable in this case) or future RV128. 

There are instances where uint64_t or uint32_t usage in the spec. But those had clear requirements.

src/sbi_cove.adoc
```diff
@@ -761,30 +769,66 @@ enum tsm_state {
...
+     * dynamic page conversion capability.
      */
-    unsigned long tvm_state_pages;
-    /* The maximum number of vCPUs a TVM can support. */
-    unsigned long tvm_max_vcpus;
+    uint64_t tvm_state_pages;
+    /* 
+     * The maximum number of vCPUs a TVM can support. 
+     */
+    uint64_t tvm_max_vcpus;
```
atishp04, line 815: same comment as below.

src/sbi_cove.adoc
```diff
@@ -761,30 +769,66 @@ enum tsm_state {
...
+     */
+    uint32_t tsm_impl_id;
+    /* 
+     * Version number of the running TSM. 
+     */
     uint32_t tsm_version;
+    /* 
+     * CoVE features supported by the running TSM.
+     * Enabled bit `i` indicates that the TSM and underly supports capability `i` and there
+     * is hardware support.
```
atishp04: The comment sounds bit ambiguous to me w.r.t "there is hardware support". I think something along the lines is sufficient.
"A bitmask of CoVE features supported by the running TSM." - We can refer to the capability table for additional clarification.

atishp04:
Apart from those nit comments, looks good to me.

src/sbi_cove.adoc
wojciechozga: Thanks, I changed these types back to `unsigned long`.

src/sbi_cove.adoc
wojciechozga: Thanks, I changed these types back to unsigned long.

src/sbi_cove.adoc
wojciechozga: The reason why I wrote it like this is that there are cases when software (TSM) and hardware support different set of features. The TSM must report to OS/VMM only features that it can use, i.e., the ones that are supported by software and hardware.

For example, take a platform with hardware support for AIA. If someone runs there a TSM that does not implement support for AIA, OS/VMM would have to use only TSM's legacy interrupt handling.

wojciechozga: I added a reference to the table as suggested.

wojciechozga:
Thanks Atish and Ravi for feedback and valuable comments. I have addressed them and pushed already changes.

rsahita (approved):
LGTM

PR #115, opened 2024-12-18
Convert `unsigned long` to `uint32_t`

### LINKED ISSUE #91: Convert `unsigned long` to `uint*_t`
The size of `long` integers are different on RV32 (4 bytes) versus RV64 (8 bytes). At least for `tsm_info.tsm_capabilities` and `tsm_info.tvm_max_vcpus` these should be fixed-width integers, as their maximum size does not depend on the RV* architecture. There may be other examples of this.

steven-bellock:
Fix #91.

### DIFF
--- a/src/sbi_cove.adoc
+++ b/src/sbi_cove.adoc
@@ -802,7 +802,7 @@ struct tsm_info {
      * `COVE_TSM_CAP_*` constants. Presence of bit `i` indicates that both the TSM
      * and hardware support the corresponding capability.
      */
-    unsigned long tsm_capabilities;
+    uint32_t tsm_capabilities;
     /*
      * The number of 4KB pages which must be donated to the TSM for storing TVM
      * state in sbi_covh_create_tvm_vcpu(). `0` if the TSM does not support the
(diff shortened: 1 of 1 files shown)


rsahita (approved):

rsahita:
@atishp04 fyi

atishp04:
tsm_capabilities is a bitmask. Thus, unsigned long allow more bits for the future for more common use case for virtualization (RV64). Currently, the bitmask table is defined accordingly as well.

Having unsigned long also brings uniformity through out the spec wherever possible.

steven-bellock:
>unsigned long allow more bits

We can make it `uint64_t`. I assume the number of capabilities will be the same for both RV32 and RV64.

src/sbi_cove.adoc
```diff
@@ -802,7 +802,7 @@ struct tsm_info {
      * `COVE_TSM_CAP_*` constants. Presence of bit `i` indicates that both the TSM
      * and hardware support the corresponding capability.
      */
-    unsigned long tsm_capabilities;
+    uint32_t tsm_capabilities;
```
rsahita, line 805:
```suggestion
    uint64_t tsm_capabilities;
```

rsahita (approved):

PR #122, opened 2025-05-13
IETF EAT is now an RFC

steven-bellock:


steven-bellock:
@rsahita @sameo

rsahita (approved):
thanks for tracking

Issue #126, opened 2025-11-06
Initial state of harts


radimkrcmar:
The specification doesn't allow VMMs to configure most architectural registers when creating a TVM.

sbi_covh_promote_to_tvm partially worked around it by allowing VMMs to set GPRs, so I think there is some interest in custom initial state, and we could make the ABI a more self-consistent, and closer to being a Virtual Machine ABI, by adding ecalls for full configuration.

Or do we want VMMs to add TVM entry code that sets the desired state?

wojciechozga:
The TSM can read CSRs and GPRs of a boot hart via NACL, but there is no way of reading other vCPUs initial state, so there is a limitation to this approach. When TSM uses a VMM-provided initial vCPU state, then initial vCPU state must be measured.

radimkrcmar:
> The TSM can read CSRs and GPRs of a boot hart via NACL,

Right, and the 1024 CSRs only cover the HS-mode CSRs so the normal S/U-mode CSRs are not expressible.

> there is no way of reading other vCPUs initial state.

We could specify that the state is an argument to vCPU creation, or must be provided by configuration ecalls between vCPU creation and finalization.
(I'd like to discuss changes to the promotion flow, so there aren't differences when they aren't important.)

> When TSM uses a VMM-provided initial vCPU state, then initial vCPU state must be measured.

Definitely.   We already must measure the VMM-provided `pc` and `a0`+`a1` GPRs, so this would expand the scope of what is measured.  (We technically measure the whole state even now.  It's just constant 0, so the work was optimized away.)

I think that defining the serialization of the whole architectural state will eventually be needed for migration, so it seems like the most viable solution to have the same structure as the input for VM/VCPU creation.
Note that the SBI state is architectural as well and will need to be fully specified.

Do we want to postpone architectural state serialization for CoVE 2.0?

Could the promotion workflow ignore NACL, and just accept `pc`, `a0`, `a1`, `hgatp`, and tap for finalization?

radimkrcmar:
We've discussed that there isn't really a need to read the data from NACL and that specifying the initial state will be sufficient for now.

@wojciechozga what other information does the single-step method use to create the TVM?
I assume we also at least need to specify the vCPU count.

wojciechozga:
**The minimum information needed to create a TVM using the promote call:**
* `pc`, `a0`, `a1`, `hgatp`, `vcpu_count`

where, 
[...]

PR #131, opened 2025-12-02
Simplify the promote call

wojciechozga:
This PR simplifies the single-step TVM creation triggered by the sbi_covh_promote_to_tvm() call by reducing the dependency on FDT and making information needed to create a TVM explicit. It also postpones the creation of a TVM from a running VM to CoVE v2, because we then expect to define the full TVM architectural state as part of the TVM migration spec.

### DIFF
--- a/src/appendix_d.adoc
+++ b/src/appendix_d.adoc
@@ -3,138 +3,101 @@
 [[appendix_d]]
 == Appendix D: M-mode TSM based deployment model
 
-This deployment model targets high-assurance systems whose design might be
-constrained
-by real-time and formal verification requirements. It trade offs a feature-rich
-design supporting
-dynamic resource allocation, for simpler implementation, where the correctness
-can be formally verified.
+This deployment model targets systems whose design is constrained by real-time
+requirements, formal verification objectives, or must operate on hardware with
+limited capabilities due to power or silicon constrains.
+It augments a feature-rich architecture with flexible ABI and dynamic resource
+management. We defined a simplified implementation that facilitates safety and
+security verification.
+Typical examples of these systems include high-end embedded or edge systems
+operating on less complex hardware than cloud-grade servers, and high-assurance
+platforms where reduced architectural complexity simplifies formal verification
+of properties such as correctness and isolation.
 
 [id=dep3]
 [caption="Figure {counter:image}"]
-[title= ": M-mode TSM based deployment model for CoVE"]
+[title= ": M-mode TSM based deployment model for CoVE."]
 image::img_11.png[align=center]
 
 === Overview
 <<dep3>> shows that the deployment model supports a single confidential
-supervisor domain in which
-the TSM runs along with the TSM driver in the M-mode. This single confidential
-supervisor domain can run multiple
-TVMs that are isolated from each other using the MMU, i.e., G-stage page tables
-managed by TSM. TSM isolates the
-hosting supervisor domain (i.e., OS/VMM and non-confidential applications and
-VMs) from the confidential supervisor
-domain (TSM and TVMs) using a hardware memory isolation mechanism, like PMP.
-The Supervisor Domain Access Protection (Smmtt) extension is therefore not
-required in this model but not precluded.
-IO accesses to confidential memory must be prevented, for example, with IOPMP.
+supervisor domain in which the TSM runs along with the TSM driver in the M-mode.
+This single confidential supervisor domain can host multiple TVMs. The TSM
+enforces isolation of TVMs using G-stage page tables. It also isolates the
+hosting supervisor domain (i.e., OS/VMM and any non-confidential applications
+and VMs) from the confidential supervisor domain (TVMs) through a hardware memory
+isolation mechanism, such as PMP. In this model, the supervisor domain access
+protection (Smmtt) extension is not required, though its use is not precluded.
+Accesses from I/O to confidential memory must be prevented, for example, though
+the use of IOPMP.
 
 [NOTE]
 ====
 Since the TSM is not required to run in the HS-mode, this deployment model
-supports systems that emulate the
-hypervisor extension or run TVMs and OS/VMM in S-mode. The latter requires use
-of a hardware memory isolation mechanism
-that enforces memory accesses to confidential memory while being only
-controlled by the TSM, e.g., PMP.
+supports systems that emulate the hypervisor extension or run TVMs and OS/VMM in
+S-mode. The latter requires use of a hardware memory isolation mechanism that
+enforces memory accesses to confidential memory while being only controlled by
+the TSM, e.g., PMP.
 ====
 
-=== Static memory partitioning
-The deployment model proposes static partitioning of memory into confidential
-and non-confidential to simplify
-formal reasoning about the correctness of the TVM execution and isolation. TSM
-performs this paritioning early
-during the boot of the platform, resulting in the following advantages: (1)
-simplified formal reasoning about the
-ownership of memory, (2) attestation that covers static system configuration
-(e.g., values of PMP registers),
-(3) reduced attack surface between OS/VMM and TSM due to narrower ABI. A
-possible negative consequence of
-static partitioning is underutilization of resources. Specifically,
-confidential memory created at platform
-initialization might be larger than the required amount of memory utilized by
-TVMs and the TSM during runtime.
-Lack of the conversion mechanism of confidential memory pages to
-non-confidential memory (enabled for example by Smmtt)
-prevents the hosting supervisor domain (OS/VMM, applications, and VMs) from
-using the memory over-provisioned by
-the confidential supervisor domain (TSM, TVMs).
+=== Static Memory Partitioning
+The deployment model should employ a static partitioning of memory into confidential
+and non-confidential regions to simplify formal reasoning about the correctness
+and isolation of TVM execution. The TSM should perform this partitioning early during
+platform boot to gain the following advantages:
+
+. simplified formal reasoning about memory ownership,
+. attestation that covers static system configuration (e.g., values of PMP
+registers),
+. reduced attack surface between OS/VMM and the TSM, achieved through
+preventing the OS/VMM from controlling the allocation of physical
+pages to TVMs and by a narrower ABI.
+
+[NOTE]
+====
+A potential drawback of static partitioning is underutilization of resources.
+Specifically, the amount of confidential memory established during platform
+may exceed the memory required by the TSM and TVMs at runtime.
+In the absence of a mechanism to convert confidential memory pages to
+non-confidential memory (as could be enabled by the Smmtt extension), the
+hosting supervisor domain (i.e., the OS/VMM, applications, and VMs) cannot
+reclaim or use memory over-provisioned to the confidential supervisor domain.
+====
 
 === TVM creation
-To reduce the complexity of the TSM implementation, the TSM creates a TVM as a
-result of a single operation triggered with
-the `sbi_covh_promote_to_tvm()` call. Specifically, OS/VMM initializes and
-starts a regular VM, which early during the
-boot process requests to be promoted to a TVM. This call traps in the OS/VMM
-and is reflected to the TSM, which copies
-VM's data, the page table configuration, and the vCPU state into confidential
-memory. If the request fails, the promotion
-of a VM to a TVM fails and error is returned to the VM. When the request
-succeeds, OS/VMM marks the VM as a TVM,
-so that it can then properly resume its execution via TSM.
+To reduce the complexity of the TSM implementation and its ABI, in this
+deployment model the TSM should create a TVM as a result of a single operation
+triggered with the sbi_covh_promote_to_tvm() call. Specifically, OS/VMM should
+constructs a regular VM in non-confidential memory and requests the TSM to
+promote it to a TVM. To do so, the TSM must copy the VM's data, page tables, and
+the boot vCPU state into confidential memory. When the request succeeds, OS/VMM
+marks the VM as a TVM, so that it can then properly resume its execution via the
+TSM.
 
-=== Local attestation
-Embedded systems might operate without access to a network, which prevents use
-of remote attestation. For this
-reason, this deployment model also supports local attestation, in which the TSM
-attests to the integrity of the TVM image
-during its creation and allows its creation only when it contains a specfic
-`TVM attestation payload` (TAP). This
-payload carries a cryptographic proof issued with the expected attestation key
-specific to the TSM integrity
-and platform configuration. The pointer to TAP is passed in a call to promote a
-VM to a TVM. If it is zero,
-then remote attestation is used, otherwise local attestation is used. Local
-attestation is strongest when it is hardware enforced.
-The TSM must verify that the attestation (i.e., integrity measurements, configuration of the hardware
-platform) of the TVM match the values specified by the TVM creator in the TAP.
+[NOTE]
+====
+The promotion mechanism enables use cases in which a running VM gets
+promoted to a TVM. To support attestation, such promotion must occur when
+the contents of VM's runtime state are predictable, allowing for reliable
+computation of reference integrity measurements. However, certain use cases
+may require promotion after that point and might even exclude some dirty VM pages
+(like those containing custom boot firmware or provisioning code) from the TVM.
+However, if your use case requires promotion after that point, then you must
+exclude dirty VM pages by modifing the FDT to mark selected pages as
+zero pages, and the TSM will clear these pages during promotion.
+====
 
-Local attestation utilizes one or more of the following properties:
-
-. The TSM must be able to verify that a TVM creator authorized the TVM to run on the platform.
-. The TSM enforces that only TVMs created by a trusted partner runs on the hardware.
-
-Separate mechanisms may be used to achieve these goals.
-
-==== TVM Authorization
-The TSM must have a list of public keys of those authorized to sign VMs (TVMs)
-for execution on the platform. The attestation payload associated with the TVM
-will be
-signed with a private key. When the VM is being promoted to a TVM, TSM checks
-the signature inside the TAP.
-If the signature is not valid, the TSM will not convert the VM and will
-terminate execution of the
-VM. The method for provisioning these public keys into the TSM is outside the
-scope of this specification.
-
-==== Verifying Platform configuration
-When the creator of the (authorized) TVM does not want it to execute on
-improperly configured or unauthorized hardware, there should be a mechanism
-supported by hardware (and firmware) for verification.
-Assuming presence of the hardware root-of-trust for measurement and hardware
-root-of-trust for storage, the VM can be created with an encrypted disk and the
-key that is used to decrypt the disk can be sealed to the measurments of the
-platform.
-The creator of the VM using the specifications for the platform decides what
-values are required in order for the key to be released.
-When the request to promote VM to a TVM is called and local attestation is
-successful, the TSM unseals the key with help of the hardware root-of-trust. At
-the point when the TVM needs to decrypt its disk (e.g., for mounting the
-filesystem), the TVM utilizes an ABI call (`sbi_covg_retrieve_secret()`) to
-retrieve the decryption key from the TSM.
-
-=== Further recommendations
-Embedded systems with real-time requirements must have a fixed upper bounded
-execution time. This requires determining
-the maximal number of instructions that can execute between TVM context
-switches. From this reason, this deployment model
-recommends an uninterruptible TSM. <<depd2>> shows this operation mode, in
-which TSM running in M-mode exposes COVH and
-COVG ABI to OS/VMM and TVM, respectively. VM ECALLs trap directly to TSM due to
-the `medeleg` configuration and all
-interrupts during TVM execution trap in TSM due to the `mideleg` configuration.
+=== Further Recommendations
+Embedded systems with real-time requirements must guarantee a fixed upper bound
+on execution time. This necessitates determining the maximum number of
+instructions that may execute between TVM context switches. For this reason,
+this deployment model recommends use of an uninterruptible TSM. <<depd2>>
+illustrates this operating mode, in which the TSM executes in M-mode and exposes
+the COVH and COVG ABIs to the OS/VMM and TVM, respectively. During the TVM
+execution, all exceptions and interrupts trap directly to the TSM, as determined
+by the medeleg and mideleg configurations.
 
 [id=depd2]
 [caption="Figure {counter:image}"]
-[title= ": TSM operation"]
+[title= ": Uninterruptible TSM operation in deployment model 3."]
 image::img_12.png[align=center]
--- a/src/appendix_d.adoc
+++ b/src/appendix_d.adoc
@@ -9,7 +9,7 @@ limited capabilities due to power or silicon constrains.
 It augments a feature-rich architecture with flexible ABI and dynamic resource
 management. We defined a simplified implementation that facilitates safety and
 security verification.
-Typical examples of these systems include high-end embedded or edge systems
+Typical examples of these systems include edge and high-end embedded systems
 operating on less complex hardware than cloud-grade servers, and high-assurance
 platforms where reduced architectural complexity simplifies formal verification
 of properties such as correctness and isolation.
@@ -47,11 +47,11 @@ and isolation of TVM execution. The TSM should perform this partitioning early d
 platform boot to gain the following advantages:
 
 . simplified formal reasoning about memory ownership,
-. attestation that covers static system configuration (e.g., values of PMP
-registers),
 . reduced attack surface between OS/VMM and the TSM, achieved through
 preventing the OS/VMM from controlling the allocation of physical
 pages to TVMs and by a narrower ABI.
+. attestation that covers static system configuration (e.g., values of some
+M-mode registers, like PMP registers),
 
 [NOTE]
 ====
@@ -67,9 +67,9 @@ reclaim or use memory over-provisioned to the confidential supervisor domain.
 === TVM creation
 To reduce the complexity of the TSM implementation and its ABI, in this
 deployment model the TSM should create a TVM as a result of a single operation
-triggered with the sbi_covh_promote_to_tvm() call. Specifically, OS/VMM should
+triggered with the `sbi_covh_promote_to_tvm()` call. Specifically, OS/VMM should
 constructs a regular VM in non-confidential memory and requests the TSM to
-promote it to a TVM. To do so, the TSM must copy the VM's data, page tables, and
+promote it to a TVM. To do so, the TSM must create the VM's data, page tables, and
 the boot vCPU state into confidential memory. When the request succeeds, OS/VMM
 marks the VM as a TVM, so that it can then properly resume its execution via the
 TSM.
@@ -82,9 +82,8 @@ the contents of VM's runtime state are predictable, allowing for reliable
 computation of reference integrity measurements. However, certain use cases
 may require promotion after that point and might even exclude some dirty VM pages
 (like those containing custom boot firmware or provisioning code) from the TVM.
-However, if your use case requires promotion after that point, then you must
-exclude dirty VM pages by modifing the FDT to mark selected pages as
-zero pages, and the TSM will clear these pages during promotion.
+However, if a use case requires promotion after that point, then dirty VM pages
+must be excluded by marking selected pages as zero pages.
 ====
 
 === Further Recommendations
--- a/src/appendix_d.adoc
+++ b/src/appendix_d.adoc
@@ -6,7 +6,7 @@
 This deployment model targets systems whose design is constrained by real-time
 requirements, formal verification objectives, or must operate on hardware with
 limited capabilities due to power or silicon constrains.
-It augments a feature-rich architecture with flexible ABI and dynamic resource
+It augments a feature-rich architecture, which offers flexible ABI and dynamic resource
 management. We defined a simplified implementation that facilitates safety and
 security verification.
 Typical examples of these systems include edge and high-end embedded systems
@@ -44,12 +44,10 @@ the TSM, e.g., PMP.
 The deployment model should employ a static partitioning of memory into confidential
 and non-confidential regions to simplify formal reasoning about the correctness
 and isolation of TVM execution. The TSM should perform this partitioning early during
-platform boot to gain the following advantages:
+platform initialzation to gain the following advantages:
 
 . simplified formal reasoning about memory ownership,
-. reduced attack surface between OS/VMM and the TSM, achieved through
-preventing the OS/VMM from controlling the allocation of physical
-pages to TVMs and by a narrower ABI.
+. reduced attack surface between OS/VMM and the TSM, achieved through a narrower ABI.
 . attestation that covers static system configuration (e.g., values of some
 M-mode registers, like PMP registers),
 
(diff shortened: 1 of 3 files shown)


wojciechozga:
fixes #126

src/sbi_cove.adoc
```diff
@@ -1117,67 +1122,100 @@ The possible error codes returned in `sbiret.error` are shown below.
...
-OS/VMM. The possible error codes are shown below.
+This function is used by the OS/VMM to promote a non-confidential VM to a TVM in
+a single operation. It is primarily intended for CoVE deployment models that
+require single-step TVM creation (e.g., <<appendix_d>>). Deployment models that
+offer multi-step TVM creation (e.g., <<dep1>>) may, but are not required to,
+support this ABI as an additional mechanism for creating TVMs.
+
+Prior to this call, the OS/VMM must set up in non-confidential memory the
+G-stage page tables, rooted at the page directory identified by `hgatp` argument
+passed via NACL shared memory, and data pages. The `hgatp` must encode the
```
radimkrcmar, line 1144: Why not pass the `hgatp` as another GPR argument, so the implementation of `sbi_covh_promote_to_tvm` doesn't need to look at the NACL structure at all?

I think there is inconsistency with passing `a1` (eintry_arg) in a GPR when there is also `a1` in the NACL.

src/sbi_cove.adoc
```diff
@@ -1117,67 +1122,100 @@ The possible error codes returned in `sbiret.error` are shown below.
...
+
+Finally, the TSM sets the TVM state to `TVM_RUNNABLE` and returns a unique TVM
+identifier (`tvm_guest_id`) in `sbiret.value`.
+
+Note that a TVM created via this call does not require a subsequent call to
+`sbi_covh_finalize_tvm()`.
+Upon successful return, the OS/VMM should release the non-confidential memory
+that contained the VM's data and page tables. Subsequently, the OS/VMM must
+interact with the TVM exclusively via the COVH ABI, resuming vCPU execution via
+`sbi_covh_run_tvm_vcpu()`. The boot vCPU must be run before any other vCPU.
```
radimkrcmar: How does the caller know the boot vCPU and why does the ordering matter?

RISC-V allows a legacy boot protocol where all harts are started at the same time and negotiate who is the boot hart among themselves.
The vCPU that is run before any other -- for an implementation-specific amount of cycles to reach a synchronization point -- becomes the boot vCPU, so there isn't really a requirement.

Without SBI HSM for CoVE, the legacy boot should actually be the only possible implementation for CoVE VMs.   (I think we should allow SBI HSM in CoVE, and specify the interaction with VMM, but that is for another PR.)

src/swlifecycle.adoc
```diff
@@ -85,21 +85,25 @@ finalized.
...
+OS/VMM requests then the TSM to promote the VM by executing the
+`sbi_covh_promote_to_tvm()` call. The TSM recreates a new TVM from the reflected
+VM state (VM data, page table configuration, vCPU state). After promotion, the
+OS/VMM should release the previously allocated VM resources and must begin
+interacting with the TVM through the TSM as described in the next sections.
+
+The mechanism by which the TSM recreates the VM state in confidential memory is
+implementation-specific. In CoVE implementations that support dynamic conversion
+between confidential and non-confidential memory, the TSM should perform page
+conversion while traversing G-stage page-table mappings. Conversely, CoVE
```
radimkrcmar: Converting in-place would go against earlier
```
Upon successful return, the OS/VMM should release the non-confidential memory
that contained the VM's data and page tables.
```
because the OS/VMM must not release (and try to then use) pages that it cannot access.

Allowing in-place conversion of hgatp pages is also a bit inefficient, because hgatp itself is in memory, so the TSM-manager would most likely to reconfigure VMM isolation and issue a global fence for each hgatp page to simplify protection against malicious VMMs.
(I find it much easier to first copy a self-consistent hgatp to VMM-inaccessible memory, make the hgatp referenced pages inaccessible from VMM, issue a global fence, and then do measurements.)

Still, we do not allow in-place conversion anywhere in CoVE so far, so I think this PR could require the copy for simplicity.

src/swlifecycle.adoc
```diff
@@ -85,21 +85,25 @@ finalized.
...
-integrity measurements calculated by TSM over the TVM data in confidential
-memory are deterministic and therefore meaningful for attestation.
-====
+and vCPUs and creating the G-stage page table configuration. This is a standard
+process for creating non-confidential VMs implemented by existing OS/VMMs. The
+OS/VMM requests then the TSM to promote the VM by executing the
+`sbi_covh_promote_to_tvm()` call. The TSM recreates a new TVM from the reflected
+VM state (VM data, page table configuration, vCPU state). After promotion, the
+OS/VMM should release the previously allocated VM resources and must begin
+interacting with the TVM through the TSM as described in the next sections.
```
radimkrcmar, line 94: The `promote` is bit of a misnomer.  A non-confidential VM cannot really be promoted, because the process loses most of the information about the VM -- the `promote` ecall still requires that the VMM built the VM specially for use with CoVE, since CoVE is *very* restricted compared to normal VMM expectations. 

Maybe it would be good to name the ecall differently?  (`sbi_covh_create_tvm_single_step`)

src/appendix_d.adoc
```diff
@@ -3,138 +3,100 @@
...
-then remote attestation is used, otherwise local attestation is used. Local
-attestation is strongest when it is hardware enforced.
-The TSM must verify that the attestation (i.e., integrity measurements, configuration of the hardware
-platform) of the TVM match the values specified by the TVM creator in the TAP.
-
-Local attestation utilizes one or more of the following properties:
+. simplified formal reasoning about memory ownership,
+. reduced attack surface between OS/VMM and the TSM, achieved through
+preventing the OS/VMM from controlling the allocation of physical
+pages to TVMs and by a narrower ABI.
```
radimkrcmar: The VMM-controlled allocation is orthogonal to static memory partitioning.

A statically partitioned system can use the multi step TVM creation method.
(And a dynamically partitioned system can only allow single step creation.)

Do we want to advise against both the `sbi_covh_convert_pages` and `sbi_covh_create_tvm` ecalls in M-mode deployments?

src/sbi_cove.adoc
wojciechozga: Good that you bring this up, both would work. The reasoning is that since NACL is required in the CoVE spec, we can make use of it to ensure backward compatibility in future versions of the spec. If future versions of the spec enable the promote call to copy the runtime state of harts, then the call implementation would only read more data from NACL, but the call's signature would remain the same. As we discussed, CoVE v2 will likely define a full architectural state of TVM in an NACL-like structure to support the migration use case. Whether of not this will be an NACL is to be discussed.

In the end of the day, we should agree on the convention and be consistent along the spec. AFAIR `sbi_covh_finalize_tvm` also passes sepc and a1 via GPRs but could via NACL.  Other thing is that there is a limited number of GRPs to pass arguments. So potential to further parametrize calls (like with a TVM-specific user-provided id) can be limited.

src/sbi_cove.adoc
wojciechozga: Right now, when we create a TVM (in a single- or multi-step way) only a boot vCPU is in a `RUNNABLE` state. A request (`sbi_covh_run_tvm_vcpu`) to run another vCPU would fail until a boot vCPU initializes it for example via SBI HSM.

`> How does the caller know the boot vCPU and why does the ordering matter?`

As till now, the (non-written) assumption for Salus and ACE was that a boot vCPU's id is 0.... 

Below sentence should be removed as the OS/VMM will currently fail running other vCPUs and if in future one can create a TVM that has a bunch of vCPUs in an initialized state, then this statement will be too strong.
```
The boot vCPU must be run before any other vCPU.
```

src/swlifecycle.adoc
wojciechozga: sounds good, let me rephrase the text.

src/swlifecycle.adoc
wojciechozga: After the cuts for the CoVE v1 the call got simplified, but the intention was to support desired call semantic in future version of the spec, once we have a way to capture the architectural state of vCPUs in NACL-like memory. Then, it should be possible to promote a non-confidential VM to a TVM similarly to how this is done in OpenPOWER PEF.

I would keep the name for now and revisit later.

src/appendix_d.adoc
wojciechozga: `> The VMM-controlled allocation is orthogonal to static memory partitioning.`

If the memory is statically partitioned, then a TSM has ownership of a memory partition. VMM cannot read/write pages within this region. How one could implement multi-step TVM creation in such a setting?

Single-step creation would work for both statically and dynamically partitioned system, since the OS/VMM provides information in non-confidential memory and the TSM can then convert these pages to confidential memory.

radimkrcmar: The CoVE ABI is fairly strictly separated into ABI for resource assignment between VMM and TSM and ABI for resource assignment between TSM and TVMs.

The TSM can allow the VMM to manage the address space even though it is statically partitioned.
If the VMM knows the address ranges that TSM uses for confidential memory (through DT, etc.), it's effectively the same as if VMM converted its own pages.  The VMM yields the ownership to TSM by performing the conversion and reclamation by VMM only works if the page is unused by TSM (which the VMM knows, since it must tell the TSM which pages to use).

For example, let's assume that addresses starting with 0x10000000 are known to be confidential and 0x00000000 are non-confidential.

The VMM could do:
```
tvm_id = sbi_covh_create_tvm(0x10000000)
sbi_covh_create_tvm_vcpu(tvm_id, 0, 0x10010000)
sbi_covh_create_tvm_vcpu(tvm_id, 1, 0x10020000)
sbi_covh_add_tvm_page_table_pages(tvm_id, 0x10030000, 0x400 - 0x30)
sbi_covh_add_tvm_measured_pages(tvm_id, 0x01234000, 0x0x10400000, 0, 0x100, 0x80000000)
sbi_covh_add_tvm_measured_pages(tvm_id, 0x05678000, 0x0x10500000, 0, 0x100, 0x80100000)
sbi_covh_add_tvm_measured_pages(tvm_id, 0x09abc000, 0x0x10600000, 0, 0x3, 0x1000000)
sbi_covh_finalize(tvm_id, 0x80000000, 0x1000000, 0x0efg0000)
```

`sbi_covh_run_tvm_vcpu(tvm_id, 0)` should work now.

And to show other aspect of the orthogonality.  If we wanted multi-step creation while forcing the TSM to allocate the confidential pages itself, we could just remove the addresses in confidential memory and do something like:
```
tvm_id = sbi_covh_create_tvm()
sbi_covh_create_tvm_vcpu(tvm_id, 0)
sbi_covh_create_tvm_vcpu(tvm_id, 1)
sbi_covh_add_tvm_measured_pages(tvm_id, 0x01234000, 0, 0x100, 0x80000000)
sbi_covh_add_tvm_measured_pages(tvm_id, 0x05678000, 0, 0x100, 0x80100000)
sbi_covh_add_tvm_measured_pages(tvm_id, 0x09abc000, 0, 0x3, 0x1000000)
sbi_covh_finalize(tvm_id, 0x80000000, 0x1000000, 0x0efg0000)
```
(This has drawbacks on dynamic reclamation, since VMM can't tell which pages are used anymore.)

src/swlifecycle.adoc
radimkrcmar: Sounds good.  Note that we'll need a new `fid` (or at least a `flag` argument) if we want to change the behavior in the future.

src/sbi_cove.adoc
radimkrcmar: > As till now, the (non-written) assumption for Salus and ACE was that a boot vCPU's id is 0....

Yeah, it's definitely the simplest to specify that TVMs use SBI HSM and 0 is the boot vCPU.

There are other aspects we should specify -- e.g. TSM must watch how VMM handled the HSM ecalls in order to keep a consistent implementation of HSM between VMM and TSM.

This funnels into a complex discussion on how to configure and handle the ecalls from TVMs...

> Below sentence should be removed as the OS/VMM will currently fail running other vCPUs and if in future one can create a TVM that has a bunch of vCPUs in an initialized state, then this statement will be too strong.

I agree.

src/sbi_cove.adoc
radimkrcmar: Maintaining the same signature and compatibility likely requires us to add a `flags` fields, so we can in the future denote to use more fields from NACL.  (Because otherwise, a new TSM working with old VMM might looks at NACL fields that weren't initialized by the VMM.)

NACL currently passes all of non-H architectural state directly, so we'd have to change its scope in order to continue using for the full architectural state.
(GPRs are non-H yet still in NACL, since we need to clobber them to invoke the ecall.
The alternative: sourcing the architectural state that isn't expressed in NACL from the architectural registers would be more to the spirit of `promote`, since that is what normal NACL VMs do, but that doesn't solve migration, because migration requires encryption and encrypted state cannot be passed in architectural registers -- e.g. WARL CSRs.)

If we endorse NACL, I would source the `a1` from there too, but `pc` is not a part of NACL.  NACL `vsepc` has different semantics from the program counter, so it doesn't seem wise to conflate them now.

Still, I'm slightly leaning against NACL when we could currently fit everything in arguments and future changes seem to require changes big enough that adding an ecall to denote that we use full serialized architectural state (possibly in encrypted form) as an input to promote/vcpu_create is fairly minor.

On a tangent: if we switch to the interfaces to MPXY, CoVE will drop NACL in favor of MPXY buffers.

src/appendix_d.adoc
wojciechozga: So OS/VMM tracks page assignments in confidential memory to correctly create/destroy TVMs while the TSM remains with the exclusive ownership and read/write access permissions to this region. Sounds good. 

I think that an OS/VMM should be able to learn which address ranges of a confidential memory are addressable, since a TSM might need some memory for its internal use. Examples of which I can think of: TSM's code/data segments, per hart dump area in the case of an interruptible TSM, internal stack, internal heap (maybe?), DICE-related information for the TSM layer, scratchpad area (maybe?), maybe some extra pages for TDISP-related structures?

src/sbi_cove.adoc
wojciechozga: `> Yeah, it's definitely the simplest to specify that TVMs use SBI HSM and 0 is the boot vCPU.`
OK, I will define that the boot vCPU has the ID 0 and for symmetric multiprocessing support inside a TVM, a TSM must support SBI HSM.

`> This funnels into a complex discussion on how to configure and handle the ecalls from TVMs...`

Absolutely! Since a TSM reflectssome of TVM's ecalls and MMIO read/write requests to an OS/VMM there is a "controlled information leakage". That is against the strict non-interference (confidentiality), but is required for any practical system. So we should be specific about the mechanism.

The security property we have been discussing for ACE is a relaxed non-interference property. The specification defines a behavior of how a TSM reclassifies information for TVM's ecalls (e.g., SBI HSM), MMIO, and shared pages. Making this explicit, we delegate to a TVM owner the decision whether or not to use the specific reclassification channels and use them in a secure way inside a TVM. 

We could think to adapt such an approach in the CoVE spec (maybe CoVE v2?). We might cover at least the standard extensions needed to support Virtio (shared pages and MMIO), symmetrical multiprocessing (SBI HSM, IPI, FENCE), virtual instructions (wfi), ... 

Some extensions, like SBI HSM, need more detailed spec. A TSM that supports SBI HSM must track a TVM's vCPU lifecycle and reflect some lifecycle changes to a OS/VMM so that OS/VMM can properly schedule vCPUs. A discussion must take place to understand if reflecting some lifecycle changes can lead to side-channels.

src/sbi_cove.adoc
wojciechozga: > Maintaining the same signature and compatibility likely requires us to add a flags fields

Shouldn't OS/VMM read a CoVE version that a TSM supports and then use the proper ABI accordingly? Or do we want to consider that a TSM can offer different versions of functions within the same implementation (I guess can end up being pretty messy)?

`> NACL vsepc has different semantics from the program counter, so it doesn't seem wise to conflate them now.`

Yes.

> Still, I'm slightly leaning against NACL when we could currently fit everything in arguments

Some other idea is to get rid of two TVM creation calls and leverage sbi_covh_create_tvm() for both single- and multi-step, just letting the caller define what type of a TVM creation is expected with a dedicated parameter. The `tvm_create_params` structure would get more complex, like a union. 

> migration requires encryption and encrypted state cannot be passed in architectural registers -- e.g. WARL CSRs.

I would expect the architectural state to be encoded in a well-defined, serialized format (NACL-like but covering the full CSR/GPR state), which is then encrypted using symmetric crypto (AEAD  like AES-GCM or ASCON). The TSM would process this in chunks: decrypt, verify integrity and versioning, and reconstruct the architectural state from the canonical representation.

Overall, migration will likely require a dedicated streaming ABI, but the underlying state format can be reused across interfaces.

src/appendix_d.adoc
wojciechozga: I fixed Appendix D.

src/swlifecycle.adoc
wojciechozga: Fixed

src/sbi_cove.adoc
wojciechozga: Fixed

src/sbi_cove.adoc
radimkrcmar: > Shouldn't OS/VMM read a CoVE version that a TSM supports and then use the proper ABI accordingly? 

VMM can't know TSM features that were ratified after the VMM was written.
If we didn't use `flags`, all we could do is to define new ecalls every time we wanted to add a feature, because existing ecalls couldn't behave differently.

VMM ideally wouldn't care about CoVE version -- I would prefer if we didn't even define it.

> Or do we want to consider that a TSM can offer different versions of functions within the same implementation (I guess can end up being pretty messy)?

Maintaining compatibility is going to be messy. :)
Instead of versions, I think it is better to think in terms of features/modules -- TSM will support various features, possibly even mandate some of them, and software will ask TSM to use a set of those features.
Older TSMs won't provide features that newer VMMs might have wanted and older VMMs won't ask for features that newer TSM might support.

The goal is to ensure that whatever usage of CoVE was correct in the past will remain correct in the future.

> Some other idea is to get rid of two TVM creation calls and leverage sbi_covh_create_tvm() for both single- and multi-step, just letting the caller define what type of a TVM creation is expected with a dedicated parameter. The tvm_create_params structure would get more complex, like a union.

Yeah, I guess we could have a signature like
```
sbi_covh_create_tvm(flags, confidential_address, vcpu_count, pc, tvm_identity_addr)
```
and the current promote would be expressed with flags:
* ALLOCATE_YOURSELF (ignores confidential_address)
* CREATE_VCPUS (creates vcpus)
* USE_HGATP (builds the guest memory from NACL hgatp, requires ALLOCATE_YOURSELF)
* FINALIZE (finalizes)

The finalize ecall would then take no arguments, since they are already provided by create.
I think it's more aligned with RISC-V design to provide composable blocks, though.
For example: keep the finalize ecall the same as now, and simplify the create_tvm signature above to be:
```
sbi_covh_create_tvm(flags, confidential_address, vcpu_count)
```
without the FINALIZE flag. The current "single-step" would be:
```
tvm_id = sbi_covh_create_tvm(ALLOCATE_YOURSELF|CREATE_VCPUS|USE_HGATP, NULL, 2)
sbi_covh_finalize(tvm_id, pc, a1, tvm_identity_addr)
```
Since hgatp is the only thing we currently read from NACL, we could just as well remove the USE_HGATP flag for a separate ecall:
```
tvm_id = sbi_covh_create_tvm(ALLOCATE_YOURSELF|CREATE_VCPUS, NULL, 2)
sbi_covh_use_hgatp(tvm_id, hgatp)
sbi_covh_finalize(tvm_id, pc, a1, tvm_identity_addr)
```
This should not be more complex to formally verify than the single-step, and it also allows a bit more flexibility -- for example to also add AIA or other future covh features before finalize.

(An alternative solution to the two create ecalls flows would be to completely get rid of the multi-step and express the whole VM configuration at once...)

src/sbi_cove.adoc
radimkrcmar: > The security property we have been discussing for ACE is a relaxed non-interference property. The specification defines a behavior of how a TSM reclassifies information for TVM's ecalls (e.g., SBI HSM), MMIO, and shared pages. Making this explicit, we delegate to a TVM owner the decision whether or not to use the specific reclassification channels and use them in a secure way inside a TVM.

I like the idea that the owner can decide what ecalls are passed to the VMM.

Do you envision that the TVM ecall configuration will be done before finalization (the configuration is explicitly added to the attestation evidence), or that the TVM will decide after it starts?

> Some extensions, like SBI HSM, need more detailed spec. A TSM that supports SBI HSM must track a TVM's vCPU lifecycle and reflect some lifecycle changes to a OS/VMM so that OS/VMM can properly schedule vCPUs. A discussion must take place to understand if reflecting some lifecycle changes can lead to side-channels.

Yeah.  If VMM has been scheduling the newly started vCPU even while it was HSM STOPPED, TSM does not need to notify VMM of anything happening in the TVM.  (high-confidentiality setup)
And if there is a need to notify VMM for scheduling reasons, there is no need to pass the `pc` that the vCPU will start at.  (Although VMM will likely already know the address, since it's likely a function of the `pc` that VMM configured.)

Similarly for IPI from SBI and IMSIC -- TSM might not need to notify VMM if they do not cooperate on guest's `wfi`.  (Well, VMM can still infer that the TVM was in `wfi` by looking at temperature, so the confidentiality gains are dubious.)

Do we want a generic start/stop scheduling vCPU notification instead of specific HSM start/stop, and IPI/wfi?

src/sbi_cove.adoc
wojciechozga: > and the current promote would be expressed with flags:

I am not convinced about the flags in create_tvm(), despite the additional flexibility they provide. The semantics become unclear, as we end up with different combinations of flags and optional parameters, some of which are invalid. This will make the TSM implementation more complex and make it harder to reason about whether the implementation is compliant with the CoVE spec.

> The finalize ecall would then take no arguments, since they are already provided by create.

This would be suboptimal for a multi-step process because the TSM would have to store information about the boot vCPU parameters from the create_tvm() call until create_vcpu() is invoked.

> An alternative solution to the two create ecalls flows would be to completely get rid of the multi-step and express the whole VM configuration at once...

We need to support multi-step TVM creation because this is required in server-class scenarios. A non-interruptible TSM that uses a single-step procedure would block a hart for the entire duration of TVM creation, which is unacceptable in server-class CoVE implementations. In some use cases, mostly embedded or edge, this may be acceptable, as TVMs there might be used to implement small, fixed functions or compartmentalization schemes rather than user-level guest OSes with gigabytes of memory.

src/sbi_cove.adoc
wojciechozga: > Do you envision that the TVM ecall configuration will be done before finalization (the configuration is explicitly added to the attestation evidence), or that the TVM will decide after it starts?

The first step is to have an explicit contract of TSM's behavior so that the owner can write a secure TVM code. A secure code would then only use reclassification channels that the owner explicitly programmed. This could be part of CoVE v1 if we really have capacity to add this.

But with emerging AI-native systems, we see that a TVM might run an adhoc generated AI code. It would be beneficial then to let the owner specify that a TVM is only allowed to use specific interfaces to prevent an unvetted code from using a channel that it is not allowed to use. A TSM should then enforce this and include these per TVM-allowed interfaces in an evidence. 

> there is no need to pass the pc that the vCPU will start at.

Exactly. When a TSM reflects HSM's hart_start() it should keep only arguments that a needed by an OS/VMM to perform its job, i.e., the vcpu id. `pc` and `opaque` arguments must not be exposed.

We do not have implemented an information flow tracking, but the idea would be to have well defined handlers inside a TSM and show that they obey to the spec. Here is how we map information coming from a confidential hart and reclassify it to an OS/VMM:
https://github.com/IBM/ACE-RISCV/blob/main/security-monitor/src/confidential_flow/handlers/symmetrical_multiprocessing/hart_start.rs

> Do we want a generic start/stop scheduling vCPU notification instead of specific HSM start/stop, and IPI/wfi?

What would be an advantage over using the existing extensions?

src/sbi_cove.adoc
radimkrcmar: > I am not convinced about the flags in create_tvm(), despite the additional flexibility they provide. The semantics become unclear, as we end up with different combinations of flags and optional parameters, some of which are invalid. This will make the TSM implementation more complex and make it harder to reason about whether the implementation is compliant with the CoVE spec.

I agree with the drawbacks of flexibility.  We definitely don't have to allow the maximal granularity.
Slightly more flexible interface can solve two issues, though:
1. The promoted TVM cannot use any additional CoVE features,
2. Multi-step creation cannot shift the memory allocation responsibilities from VMM to TSM.

I think the first issue is quite important in the long term, since we don't want to integrate every new CoVE feature twice -- for both multi and single step creation.  (For example, imagine that we want to properly define the ecall interfaces before finalization as discussed in the other thread.)

Would ACE be fine with a two or three step creation?

```
tvm_id = sbi_covh_create_tvm_allocate(2)
sbi_covh_use_hgatp(tvm_id, hgatp)
sbi_covh_finalize(tvm_id, pc, a1, tvm_identity_addr)
```
(I've now encoded the flags in fid to constrain the flexibility, hgatp could be folded in to achieve two-step.)

> This would be suboptimal for a multi-step process because the TSM would have to store information about the boot vCPU parameters from the create_tvm() call until create_vcpu() is invoked.

Yes, the 2*XLEN bits of storage would probably be wasted after finalization, but it simplified the ABI.

> We need to support multi-step TVM creation because this is required in server-class scenarios. A non-interruptible TSM that uses a single-step procedure would block a hart for the entire duration of TVM creation, which is unacceptable in server-class CoVE implementations.

Right, we would have to figure out how to make the ecall interruptible, which seems even more complicated to get right... and RISC-V ecosystem prefers composable operations.

src/sbi_cove.adoc
radimkrcmar: >The first step is to have an explicit contract of TSM's behavior so that the owner can write a secure TVM code. A secure code would then only use reclassification channels that the owner explicitly programmed. This could be part of CoVE v1 if we really have capacity to add this.

I think that writing a contract of TSM's behavior and of the execution environment provided to TVM is the purpose of CoVE.  All of this is technically a function of TSM's code, hence a part of TVM evidence, but I'd at least try to prevent vastly different confidentiality guarantees from various CoVE implementations, since specifying something that never behaves the same in important aspects is a bit questionable...

We'll likely find out that we need to add programmability for crucial behaviors, like wfi.
If we don't manage to do that for CoVE 1.0, then we might be required to specify the most confidential solution (e.g. wfi never returns to VMM) and leave room for future extensions that loosen it.

How much confidentiality does ACE trade to improve the practicality?

>But with emerging AI-native systems, we see that a TVM might run an adhoc generated AI code. It would be beneficial then to let the owner specify that a TVM is only allowed to use specific interfaces to prevent an unvetted code from using a channel that it is not allowed to use. A TSM should then enforce this and include these per TVM-allowed interfaces in an evidence.

Protecting TVM against itself doesn't fit in the scope of trusted/confidential computing.  It's good if what we need for the programmable explicit contract also allows this use-case, but I wouldn't force it.

> https://github.com/IBM/ACE-RISCV/blob/main/security-monitor/src/confidential_flow/handlers/symmetrical_multiprocessing/hart_start.rs

I think the VMM might want to decide whether the HSM start ecall succeeds or not -- if the OS has no intention of scheduling the VM (perhaps because you didn't pay for the extra vCPUs), then TVM's ecall would ideally fail, but the TSM can't know that before returning to VMM and getting its response...
If ACE does not actually wait for VMM's response, wouldn't it be fine to reclassify that non-boot vCPUs start as not-being-scheduled, and VMM only receives notification that it should start scheduling a given vCPU?

>> Do we want a generic start/stop scheduling vCPU notification instead of specific HSM start/stop, and IPI/wfi?
> What would be an advantage over using the existing extensions?

It's probably fine to keep the distinction between HSM start/stop and IPI/wfi/suspend in VMM notifications at the cost of confidentiality, but distinguishing various "interrupt" methods -- IMSIC IPI, SBI IPI, SBI pv_kick -- and various halts -- wfi, SBI suspend, SBI pv_halt -- doesn't seem to increase practicality and only loses confidentiality.

I would like if we specified what notifications to start/stop scheduling are delivered to VMM and coerced as many sources as possible to the same notification.

src/sbi_cove.adoc
wojciechozga: I agree with you on a general direction and I recognize that the discussion is not only about the promote vs multi-step creation processes. The current spec is now about two distinct models, each coming with its sets of features, but it should rather define a set of features that can be combined arbitrarily. A company who has a use case should decide what features it needs. Our job is to define a clean ABI to support this.

Current spec defines two models: a feature-rich model that targets datacenter-class use cases and (2) a slimmed-down model for embedded/edge systems (see the table at the end). The problem is that each model comes with a set of features that could be used also in the other model but the ABI is not inclusive enough to enable that, e.g., DM1 cannot authorize if a TVM can execute on the system and DM3 cannot support directly-injected interrupts. A flexible ABI would rather enable systems that implement any combination of these features. These features (selected during the design of a CoVE implementation) are:

* Memory partitioning: Static | Dynamic
* Authorized execution: Yes | No
* Attestation and secret delivery: Local | Remote
* Page allocator: OS/VMM | TSM
* TVM memory layout: HGATP | Multi-step
* Interrupt handler: Emulated (A/PLIC) | Directly-injected (AIA IMSIC) 
* Timers: TSM-Emulated | Hardware-virtualized (RISC-V Sstc)

On top of that, there comes a list of (SBI) extensions that a TSM can offer to a TVM (a TVM might use them or not)
* Multi-vCPU support (SBI HSM, IPI): Yes | No
* IO support: Console | VirtIO (Shared Pages, VMM-exposed MMIO load/store) | CoVE-IO
* WFI: Yes | No

Our current deployment models would look like:

| Feature | DM1 | DM2 | DM3 |
| -------- | -------- | -------- | -------- |
| Memory partitioning | Dynamic (Smmpt) | Dynamic (nested virt.) | Static |
| Authorized execution | No | No | Yes |
| Attestation and secret delivery | Remote | Remote | Local |
| Page allocator | OS/VMM | OS/VMM | TSM |
| TVM memory layout | Multi-step | Multi-step | HGATP | 
| Interrupt handler | Directly-injected | Directly-injected | Emulated | 
| Multi-vCPU | Yes | Yes | Yes | 
| IO Support | VirtIO | VirtIO | VirtIO | 
| WFI |  |  | No | 

A company, which implements CoVE for a specific use case, decides on features to implement to balance between TCB, performance, silicon, power.

src/sbi_cove.adoc
wojciechozga: >  specifying something that never behaves the same in important aspects is a bit questionable...

agreed

> I think the VMM might want to decide whether the HSM start ecall succeeds or not

We should define it and ACE (as well as other implementations) must follow what we agree on.

We actually treated `HSM::start()` as a way to inform an OS/VMM that a vCPU started and OS/VMM should start to schedule it. But as you pointed out, `HSM::start()` can be interpreted differently: a TVM requests OS/VMM to start a vCPU and OS/VMM decides based on some licensing data whether or not to allow that. We should check th spec 
` on a TVM's HSM::start() request, a TSM might move a vCPU to `START_PENDING` and then on reply form OS/VMM, move to STARTED or 

 We should then stick to one version that is the closest to the HSM::start() semantics [HSM spec](https://github.com/riscv-non-isa/riscv-sbi-doc/blob/master/src/ext-hsm.adoc) [HSM state machine](https://github.com/riscv-non-isa/riscv-sbi-doc/blob/master/images/riscv-sbi-hsm.png) and define the expected behavior in the CoVE spec. 


> I would like if we specified what notifications to start/stop scheduling are delivered to VMM and coerced as many sources as possible to the same notification.

agreed

src/sbi_cove.adoc
wojciechozga: > Would ACE be fine with a two or three step creation?

Yes, as long as we keep some form of copying the VM state by traversing over the page table hierarchy (e.g., via `sbi_covh_use_hgatp()`) then we are fine

PR #142, opened 2026-05-26
sbi_cove: initial version of memory mappings

radimkrcmar:
Replace the EMTT by explicit memory tracking state machines that are integrated as pre-conditions and post-conditions to the ecalls.

The HPA diagram renders as:
image

The GPA state machine is (I think unnecessarily) complicated because of the invalidation protocol, and it already forbade some flows like interleaving of host side invalidation and guest side sharing/unsharing.

The GPA diagram renders as:
image

### DIFF
--- a/src/sbi_cove.adoc
+++ b/src/sbi_cove.adoc
@@ -138,9 +138,148 @@ partitioning of confidential and non-confidential memory (for example
 execution of untrusted code, and thus do not need to implement above ABI.
 ====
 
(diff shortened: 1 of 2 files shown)


radimkrcmar:
I added a dirty preview of where I was leading with the first patch in case we have time to discuss it on the meeting.
The idea is to simplify the GPA diagram into this by making the sharing less complicated: 
image

